PR #28, opened 2012-11-11
Added support to specify custom http headers

luk-:
There have been a couple situations where I'm using this module and wishing I could quickly specify custom http headers in the options argument. This allows that. Added a test for it, existing tests pass.


jycabrera0628:
I am trying to add csp to my website, for this I am using http-server, how can I create custom http header?

PR #60, opened 2013-12-13
Added useful tip with local addresses

smileart:
Added useful tip with local addresses not to run ifconfig every time to remember your local IP.


indexzero:
I do really want to accept your work @smileart but things have diverged too much. If you'd like to resubmit I would very much appreciate it, but I understand if you're not interested.

PR #64, opened 2014-01-09
accept command line arg for a custom header

skivvies:
This builds on @luk-'s work in #28 to allow specifying a custom header via the -H or --header command line argument.

A future change could allow for specifying more than one custom header via a --headers command line argument. It looks like optimist does not support recovering multiple passed-in -H arguments, so unfortunately emulating curl isn't an option (e.g. `curl -H 'X-Foo: Bar' -H 'X-Baz: Qux' http://www.google.com/`).

For now, this is definitely useful for e.g. adding an [HSTS header](https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security), so I thought I'd send a pull request.


indexzero:
Closing. No follow-up from author.

PR #88, opened 2014-09-05
Get CORS support working with XHR preflight

paulmelnikow:
Without this patch, http-server returns 404 on the preflight request (method: 'OPTIONS'). You can see the problem by loading an XHR request in Chrome.

### DIFF
--- a/bin/http-server
+++ b/bin/http-server
@@ -70,10 +70,7 @@ function listen(port) {
   };
 
   if (argv.cors) {
-    options.headers = {
-      'Access-Control-Allow-Origin': '*',
-      'Access-Control-Allow-Headers': 'Origin, X-Requested-With, Content-Type, Accept'
-    };
+    options.cors = true;
   }
 
   if (ssl) {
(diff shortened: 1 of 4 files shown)


paulmelnikow:
ping @indexzero, can you get this merged?

indexzero:
@paulmelnikow can you rebase off `master`?

paulmelnikow:
done

indexzero:
@paulmelnikow read a little closer, there is actually a `--cors` option in the CLI, but in this implementation the `corser` middleware is added by default. Can you make it respond to the option? Then I'll merge. 

Bonus points if you write tests.

paulmelnikow:
Ah, makes sense. I'll take another pass.

paulmelnikow:
Updated, now honoring the cors option. Includes a test which fails if you remove `cors: true`. Moved the cors header config into http-server.js so it works in the test.

indexzero:
Awesome! Merging. Thanks for writing the tests :100:

PR #117, opened 2014-12-13
New feature about display my external ip

Jinkwon:
I made some simple code with ip module.
This is simple idea.
When I work with http-server, many time I forgot ip address for mobile web browser test.
The example here below.


indexzero:
I do really want to accept your work @Jinkwon but things have diverged too much. If you'd like to resubmit I would very much appreciate it, but I understand if you're not interested.

PR #125, opened 2015-02-04
Added logging for errors (e.g. 404 - file not found)

eladmoshe:
Currently, when requests fail, there is no indication for that in the console. This code change prints an error message. For example:
[Wed, 04 Feb 2015 21:23:23 GMT] "GET /dummy" Error (404): "Not found"


indexzero:
Two space soft tabs please. On the plus side I found out that our TravisCI integration was broken today.

indexzero:
Cherry-picked. Thanks for updating the style!

PR #129, opened 2015-02-21
Change default host to 127.0.0.1

elvinyung:
`0.0.0.0` is non-routable, and non-visitable in Chrome.

`127.0.0.1` is a better default host.

### DIFF
--- a/bin/http-server
+++ b/bin/http-server
@@ -14,7 +14,7 @@ if (argv.h || argv.help) {
     "",
     "options:",
     "  -p                 Port to use [8080]",
-    "  -a                 Address to use [0.0.0.0]",
+    "  -a                 Address to use [127.0.0.1]",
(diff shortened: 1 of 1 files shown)


rayshan:
+1 chrome defaults to a google search with http://0.0.0.0:8080

impinball:
+1 for the same reason.

Ping @indexzero - this is a bit of an annoyance. It would be easier than my usual command line now:

```
hs -a localhost
```

Please merge this ASAP

yfr:
+1 for the same reason. (copy&paste commenting)

indexzero:
The approach in #132 is better. The beef here is with the log message and not actually binding to `0.0.0.0`. In other words, one could still manually type "127.0.0.1:8080" into Chrome and it would go to the page served by `hs`.

PR #130, opened 2015-02-23
Forwarding through reverse proxy for non-static files

arvind-agarwal:
I have added support for basic reverse proxy. The way it works is if file is not found by ecstatic, it will fall back to forwarding the request to a proxy server, configured by -f parameter, for .e.g  http-server -f http://localhost:5000/ etc. This is very basic support to keep in line with spirit of http-server to keep it simple and cli driven. 

We are using this as small replacement of nginx in dev environment. So that static files can be served as well as it works as reverse proxy for rest of the requests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,8 @@ This will install `http-server` globally so that it may be run from the command 
 
 `-K` or `--key` Path to ssl key file (default: key.pem).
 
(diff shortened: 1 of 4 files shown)


indexzero:
Thank you for your contribution. This ended up being a duplicate of #134. In the future conforming to the project code style is a big :+1:

Issue #67, opened 2014-02-25
Add support for a configuration file (.httpserverrc?)


lukaszfiszer:
I use http-server to develop multiple projects on my local machine. I want each project to be launched from a different port - so that I can use Chrome DevTools' Workspaces for editing with no problems (DevTools does "per hostname" binding between remote files and local files).

Currently http-server does not allow to load configuration options from a file so I need to manually provide and remember each port for each of my projects. 

My proposition is to have a .httpserverrc file (like .bowerrc or .jshintrc) which will allow to store the configuration in a JSON format.

Related issue: #24

PR #131, opened 2015-02-25
Replace optimist with nconf and look for .httpserverrc settings file

### LINKED ISSUE #24: Use nconf for configuration management
We should replace the current argument parsing with `optimist` and replace it with `nconf`.

This will allow `http-server` to be configured via the command line, or a `config.json` file, or any other transport `nconf` supports.

timothykang:
I thought JSHint's method of searching up the directory tree would work nicely, as it would allow projects to share settings while also allowing overrides.

Related to #67, closes #24.

### DIFF
--- a/bin/http-server
+++ b/bin/http-server
@@ -8,13 +8,15 @@ var colors     = require('colors/safe'),
     portfinder = require('portfinder'),
     opener     = require('opener'),
     fs         = require('fs'),
-    config     = require('nconf').argv({
+    nconf      = require('nconf'),
+    path       = require('path'),
+    pjson      = require('../package.json');
+
+var config = nconf.argv({
       cors: { boolean: true },
       'log-ip': { boolean: true }
     }),
-    pjson      = require('../package.json');
-
-var ifaces = os.networkInterfaces();
+    ifaces = os.networkInterfaces();
 
 process.title = 'http-server';
 
@@ -61,6 +63,24 @@ if (config.get('h') || config.get('help')) {
   ].join('\n'));
   process.exit();
 }
+
+var root = config.get('_')[0];
+
+if (!root) {
+  try {
+    fs.lstatSync('./public');
+    root = './public';
+  }
+  catch (err) {
+    root = './';
+  }
+}
+
+config.file({
+  file: '.httpserverrc',
+  dir: path.resolve(root),
+  search: true
+});
 
 var port = config.get('p') || config.get('port') || parseInt(process.env.PORT, 10),
     host = config.get('a') || '0.0.0.0',
--- a/bin/http-server
+++ b/bin/http-server
@@ -8,17 +8,17 @@ var colors     = require('colors/safe'),
     portfinder = require('portfinder'),
     opener     = require('opener'),
     fs         = require('fs'),
-    argv       = require('optimist')
-      .boolean('cors')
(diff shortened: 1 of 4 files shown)


indexzero:
Cool idea! Let me review this weekend.

bin/http-server
```diff
@@ -4,11 +4,16 @@ var colors = require('colors'),
     httpServer = require('../lib/http-server'),
     portfinder = require('portfinder'),
     opener = require('opener'),
-    argv = require('optimist')
-      .boolean('cors')
-      .argv;
+    fs = require('fs'),
+    path = require('path'),
+    nconf = require('nconf')
```
indexzero: Some people use this as a library. We should really be doing:

``` js
var nconf = require('nconf');

var config = new nconf.Provider()
  .argv({
    d: { boolean: true, default: true },
    i: { boolean: true, default: true },
    cors: { boolean: true }
  });
```

package.json
```diff
@@ -59,9 +59,9 @@
   ],
   "dependencies": {
     "colors": "1.0.3",
-    "optimist": "0.6.x",
     "union": "~0.4.3",
     "ecstatic": "~0.6.0",
+    "nconf": "^0.7.1",
```
indexzero: Just released `nconf@0.8.x` this past weekend. Lets upgrade to that.

indexzero:
+1 if you can rebase of `master`

timothykang:
Rebased and addressed comments. In the process I undid a small change I made to how the boolean `showDir` and `autoIndex` options were being parsed that would've broken backwards compatibility.

indexzero:
Will take a look at this over the weekend or next week.

timothykang:
Rebased onto latest `master`.

timothykang:
Noticed the recent release, rebased onto `master` again.

kilianc:
+1

timothykang:
An oldie indeed. Rebased onto current master, in light of the renewed interest.

thornjad:
@timothykang I worked on replacing optimist with minimist in #464, which I still believe is a step in the right direction, but I have to say I like this nconf approach better! 

Would you mind taking a look at the merge confict in bin/http-server? Could you also look into specifying the current version of nconf, and make sure it doesn't break anything?

Thank you!

timothykang:
@thornjad Done!

bin/http-server
```diff
@@ -45,14 +46,32 @@ if (argv.h || argv.help) {
...
+
+config.file({
+  file: '.httpserverrc',
+  dir: path.resolve(root),
+  search: true
+});
+
+var port = config.get('p') || config.get('port') || parseInt(process.env.PORT, 10),
+    host = config.get('a') || '0.0.0.0',
+    ssl = !!config.get('S') || !!config.get('ssl'),
```
thornjad: I _think_ we can use `parseValues: true` in `.argv` above rather than need to coerce here. Correct me if I'm wrong on that, I haven't used nconf before.

timothykang:
@thornjad Rebased. I added value parsing for numbers and where it made sense, but I left the keys as the short forms since not every option had a long form (they could be easily added in a follow-on PR).

karlhorky:
I guess this can probably be closed, now that #622 was merged instead (switch `optimist` to `minimist`)

PR #132, opened 2015-03-01
Use 127.0.0.1 when listening on 0.0.0.0

1j01:
Resolves https://github.com/nodeapps/http-server/issues/128
Contrast to https://github.com/nodeapps/http-server/pull/129
Instead of changing the default to 127.0.0.1, this simply opens 127.0.0.1 as the canonical url for 0.0.0.0


indexzero:
Cherry-picked. Thanks!

PR #141, opened 2015-03-20
Added useful tip with local addresses

smileart:
Added useful tip with local addresses not to run `ifconfig` command every time you want to recall your local IP.
Once again :wink: #60

### DIFF
--- a/bin/http-server
+++ b/bin/http-server
@@ -5,6 +5,8 @@ var colors = require('colors'),
     argv = require('optimist').argv,
     portfinder = require('portfinder'),
     opener = require('opener');
+    os = require('os');
+    ifaces = os.networkInterfaces();
 
 if (argv.h || argv.help) {
   console.log([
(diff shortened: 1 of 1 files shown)


indexzero:
lgtm at first glace. I will double check over the weekend and (likely) merge this in as is.

smileart:
Looks like it's going to repeat its predecessor's fate which had awaited 1 year 3 month and 7 days to be ~~not~~ merged :grinning: :octocat:

bin/http-server
```diff
@@ -98,6 +101,16 @@ function listen(port) {
       + ' on: '.yellow
       + (protocol + '//' + host + ':' + port).cyan);
 
+    log('\nAvailable on:'.yellow);
+
+    for (var dev in ifaces) {
```
indexzero: Do not use `for in` loops.

indexzero:
Some comments for you.

bin/http-server
indexzero: Use `Object.keys(ifaces).forEach(function (dev) {` instead.

bin/http-server
```diff
@@ -98,6 +101,16 @@ function listen(port) {
...
+      ifaces[dev].forEach(function(details){
```
indexzero: Conform to style here: `ifaces[dev].forEach(function (details) {`

smileart:
Are there still any chances of merging or at least adding such a simple function? I'm just curious, nothing personal. :innocent:

indexzero:
Whoops. Didn't get emails for your new commits. In the future comment after committing and I'll be sure to merge it faster.

PR #150, opened 2015-04-16
Cors with additional headers

MichalCz:
I added a small mod that allows to pass additional allowed request headers to corser.

This is done by passing: --corsheaders="Header,List,Separated,By,Commas".


bin/http-server
```diff
@@ -80,6 +81,10 @@ function listen(port) {
     options.cors = true;
   }
 
+  if (argv.corsheaders) {
+    options.cors_headers = argv.corsheaders;
```
indexzero: `options.corsHeaders`.

bin/http-server
```diff
@@ -20,6 +20,7 @@ if (argv.h || argv.help) {
     "  -e --ext           Default file extension if none supplied [none]",
     "  -s --silent        Suppress log messages from output",
     "  --cors             Enable CORS via the 'Access-Control-Allow-Origin' header",
+    "  --corsheaders      Additional CORS headers",
```
indexzero: This doesn't need to be a separate option. We should simply expect `--cors` to also be a string sometimes.

indexzero:
@MichalCz will accept this once:

1. Tests are passing (mostly style it seems)
2. There are tests for this feature
3. It is implemented as a single option (e.g. `--cors` or `--cors "Header,List,Separated,By,Commas"`). No need for `--corsheaders`.

MichalCz:
Will fix next week. :)

MichalCz:
Hey @indexzero 

I rebased the changes, added tests, fixed changes. If you had any other requests let me know.

BTW - I still use a sepatate option in the javascript interface so that there's no need to check type in the http-server.js code, I changed command line usage only.

/Mike

MichalCz:
Oh, wait - one more fix to do...

/Mike

MichalCz:
@indexzero - code should now comply with what you wanted. :)

/Mike

indexzero:
Thanks!

danielweck:
Nice one!
Fixes https://github.com/indexzero/http-server/issues/154

Issue #154, opened 2015-05-18
HTTP headers overwrite user config, just before CORS preflight


danielweck:
https://github.com/indexzero/http-server/blob/master/lib/http-server.js#L46

```javascript
if (options.cors) {
    this.headers['Access-Control-Allow-Origin'] = '*';
    this.headers['Access-Control-Allow-Headers'] = 'Origin, X-Requested-With, Content-Type, Accept';

    before.push(corser.create());
}
```

For example, in our app we need the `Range` header in `Access-Control-Allow-Headers`, but the above code overwrites our command line / JSON configuration.

Suggested fix: remove the predefined HTTP headers, which should be at the discretion of client-side usage, not baked / hard-coded as default values:

```javascript
    this.headers['Access-Control-Allow-Origin'] = '*';
    this.headers['Access-Control-Allow-Headers'] = 'Origin, X-Requested-With, Content-Type, Accept';
```

danielweck:
Related: https://github.com/indexzero/http-server/issues/145

StyMaar:
I'm not sure I understand your proposal. 
“Suggested fix: remove the predefined HTTP headers, which should be at the discretion of client-side usage, not baked / hard-coded as default values:”
Do you want to replace the `--cors` option by user set `'Access-Control-Allow-Headers` ? 
What “JSON configuration are you talking about ” ?

StyMaar:
Oh, maybe you're referring to this pull request ? https://github.com/indexzero/http-server/pull/64

danielweck:
Both `Access-Control-Allow-Origin` and `Access-Control-Allow-Headers` should be totally configurable on the client side of the HTTP-Server API. I am not against sensible defaults, but the current implementation does not even allow us ; i.e. users of the HTTP-Server library / utility ; to alter (remove, add) these predefined headers, which are forced to arbitrary values despite the passed user-defined `options.headers`:

https://github.com/indexzero/http-server/blob/master/lib/http-server.js#L24
```javascript
 this.headers = options.headers || {};
```

https://github.com/indexzero/http-server/blob/master/lib/http-server.js#L46
```javascript
if (options.cors) {
    this.headers['Access-Control-Allow-Origin'] = '*';
    this.headers['Access-Control-Allow-Headers'] = 'Origin, X-Requested-With, Content-Type, Accept';

    before.push(corser.create());
  }
```

danielweck:
[...]

PR #157, opened 2015-06-10
Allow byte range requests over CORS Fixes #156

### LINKED ISSUE #156: Access-Control-Allow-Headers does not allow Range
Hi there, big fan of http-server.  I'm trying to use it to serve content locally for testing MSE with DASH which makes heavy use of HTTP Byte Range Requests.  It look like blink based browsers complain that "XMLHttpRequest cannot load http://localhost:8000/audio_128k.webm. Request header field Range is not allowed by Access-Control-Allow-Headers."

I run http-server with the --cors flag.  The DASH libraries I'm testing both seem to send HTTP OPTIONS requests to the assets to check the response headers (preflight).  It looks like http-server is responding with the following headers: "Access-Control-Allow-Headers:Origin, X-Requested-With, Content-Type, Accept"

So I think Range would have to be added to the list.  I'm curious, is there a reason why not all headers are whitelisted with the asterisk as a value?

nickdesaulniers:


nickdesaulniers:
This works for me, not sure if you need a unit test, or would prefer the asterisk as the value.

lib/http-server.js
```diff
@@ -45,7 +45,7 @@ var HTTPServer = exports.HTTPServer = function (options) {
 
   if (options.cors) {
     this.headers['Access-Control-Allow-Origin'] = '*';
-    this.headers['Access-Control-Allow-Headers'] = 'Origin, X-Requested-With, Content-Type, Accept';
+    this.headers['Access-Control-Allow-Headers'] = 'Origin, X-Requested-With, Content-Type, Accept, Range';
```
danielweck: Sensible defaults are a good thing, but this code *overrides* any potential specified `options.headers`, instead of merging them. See:

https://github.com/indexzero/http-server/issues/154#issuecomment-106784449

indexzero: @danielweck a PR for this would be welcomed.

danielweck: I added some info in #154

danielweck: Issue addressed here:
https://github.com/indexzero/http-server/pull/150

PR #159, opened 2015-06-12
Allows explicitly setting cache to 0, closes #158

### LINKED ISSUE #158: setting `options.cache` to 0 causes the default 3600 to take precedence
This is caused directly by line 26 in `lib/http-server.js` which is:

    this.cache = options.cache || 3600; // in seconds.

Should be:

    this.cache = options.cache === undefined ? 3600 : options.cache

To allow explicitly setting `cache` to 0.

karolmajta:
This closes https://github.com/indexzero/http-server/issues/158

Should be ok to merge. I strongly feel this is the desired behavior - in normal circumstances you should be able to explicitly set `max-age` to zero.

### DIFF
--- a/lib/http-server.js
+++ b/lib/http-server.js
@@ -23,7 +23,7 @@ var HTTPServer = exports.HTTPServer = function (options) {
 
   this.headers = options.headers || {};
 
-  this.cache = options.cache || 3600; // in seconds.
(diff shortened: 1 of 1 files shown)


burczyk:
It took me a while today to figure out why my changes are not visible in the browser. Thanks for that!

You've got my +1

Morgaroth:
+1

PR #165, opened 2015-07-17
Add option to output only error messages

kettuniko:
Thought that this could come in handy for others as well.
Better parameter name suggestions welcome :)


indexzero:
We should handle this holistically with a `loglevel` handler.

PR #166, opened 2015-07-21
add livereload

tamamadesu:
add livereload module 

How to work ?

First, install the LiveReload browser plugins by visiting http://help.livereload.com/kb/general-use/browser-extensions. and:
> $  http-server -w

### DIFF
--- a/bin/http-server
+++ b/bin/http-server
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
-
+"use strict";
 var colors     = require('colors'),
     os         = require('os'),
     httpServer = require('../lib/http-server'),
@@ -44,6 +44,7 @@ var port = argv.p || parseInt(process.env.PORT, 10),
     log = (argv.s || argv.silent) ? (function () {}) : console.log,
     ssl = !!argv.S || !!argv.ssl,
     proxy = argv.P || argv.proxy,
+    livereload =  argv.w || false,
     requestLogger;
 
 if (!argv.s && !argv.silent) {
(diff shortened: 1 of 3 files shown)


indexzero:
@tamamadesu thank you for your contribution. Your PR is not passing tests because it fails lint. Try running `npm test` and making the changes. The list of errors is [also in the TravisCI job](https://travis-ci.org/indexzero/http-server/jobs/71944802)

indexzero:
@tamamadesu cool feature! I'm going to wait until more people request it before considering adding it in.

evanplaice:
:+1: @indexzero consider me added to the request list

Express and [insert server middleware] is overkill for devs working on SPAs that don't require a data back-end, or pull data from an api client-side via AJAX.

Http-server is perfect for developing SPAs, livereload support is the last missing piece preventing me from ditching express on projects where I don't need to server-side app.

piotrwitek:
+1, I'm feeling the same here, the last missing piece, for now just using alternatives

evanplaice:
@piotrwitek Take a look at live-server. It basically works the same as http-server but provides live-reload and open-browser by default.

piotrwitek:
@evanplaice thanks a lot for your suggestion :), actually I'm using jspm-server as an alternative, a fork of live-server.
Still I think that livereload is now kinda standard, and should be implemented anyway to keep up with the competition :)

gtcdevop:
@indexzero I would appreciate that feature man

PR #169, opened 2015-08-04
Handle sigterm to exit

francois2metz:
I run http-server in a docker container.

When docker stop a container, it send a SIGTERM, and after a timeout send a SIGKILL.

By default, node && iojs trap sigterm and does nothing. This trigger the timeout on docker to send a sigkill. 
So basically, stopping a container running `http-server` take some seconds.

Now, it's almost instantaneous. :boom:

### DIFF
--- a/bin/http-server
+++ b/bin/http-server
@@ -138,3 +138,8 @@ process.on('SIGINT', function() {
   log('http-server stopped.'.red);
   process.exit();
 });
+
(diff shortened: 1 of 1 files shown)


ronag:
This would be nice... Any reason it's not been merged?

francois2metz:
@ronag It was merged more than 6 month ago. \o/

PR #171, opened 2015-08-07
Add -u option to open browser with custom url

hugomn:
In some cases you have a aliased domain for your localhost (e.g: local.mydomain.com) it's necessary you open the browser after running the http-server pointing to that domain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,9 @@ This will install `http-server` globally so that it may be run from the command 
 
 `--cors` Enable CORS via the `Access-Control-Allow-Origin` header
 
-`-o` Open browser window after staring the server
+`-o` Open browser window after starting the server
+
+`-u` Url to display in the opened browser (in case you have an alias to your localhost)
 
 `-c` Set cache time (in seconds) for cache-control max-age header, e.g. -c10 for 10 seconds (defaults to '3600'). To disable caching, use -c-1.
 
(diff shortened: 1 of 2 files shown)


indexzero:
Could you also update the README.md?

hugomn:
Done! =)

hugomn:
Any news on this?

BigBlueHat:
This seems useful, but we may need to change the parameter from `-u` to `--url` after some of the other things like #245 get addressed--to be sure we're not going to make things too confusing parameter-wise. It's queued for v0.10.0 though.

README.md
```diff
@@ -60,7 +60,9 @@ This will install `http-server` globally so that it may be run from the command
 
 `--cors` Enable CORS via the `Access-Control-Allow-Origin` header
 
-`-o` Open browser window after staring the server
+`-o` Open browser window after starting the server
+
+`-u` Url to display in the opened browser (in case you have an alias to your localhost)
```
thornjad: could we actually change this to `--url` to avoid confusion with `--utc` and `--username`?

thornjad:
Actually @hugomn, can this functionality be achieved with the `-o` option?

README.md
hugomn: No, it's not possible because `-o` doesn't let you specify the desired url to open. I'll check the conflicts (3 years after!).

thornjad:
> No, it's not possible

What about using `-a`? Like with the URL in your first comment, could you do this with `-a local.mydomain.com -o /my/path`?

> (3 years after!)

Yeah I know, we're working on it :)

thornjad:
Closing due to inactivity, re-open if we'd still like to get this functionality. In my opinion, this would be better added as part of `-o` rather than as its own option.

PR #175, opened 2015-09-04
Added option to disallow all 3rd party request using CSP

2vek:
This PR adds a new option `--no3p` which uses `Content-Security-Policy` header to disallow requests to 3rd party domains(for fonts, styles, scripts etc). 
Content Security Policy is [well supported](http://caniuse.com/#feat=contentsecuritypolicy).

**Why?**
Offline docs of some projects I have downloaded,  make 3rd party request for google font, some scripts even analytics. Aside from privacy issue, every time I navigate between pages, they generate lots of requests.


2vek:
Closing the pull request as I have not used this feature myself in past 6 months or so. This pull request was trivial and only add CSP header with value `default-src 'self'`.

PR #188, opened 2015-10-06
live reload support

jso0:
live reload,that's better


thornjad:
I'm not sure this is in-scope for `http-server`. @BigBlueHat what do you think?

BigBlueHat:
Agreed. This isn't what `http-server` is currently targeting. It's meant to be more a fire-and-forget-it-from-the-command-line sort of tool.

We're also keen to avoid scope creep right now while we work to solidify the current code and move it toward the future (since someone keeps moving the ground underneath all that is JavaScript...).

Thanks for the contribution, though, @jso0! Sorry it's not quite a fit.

PR #191, opened 2015-10-08
Fix non-showing double-slashes at the end of the protocol

gokaygurcan:
On MacOS and Windows, I experienced that there're missing slashes like this:

```
Starting up http-server, serving ./
Available on:
  [public ips are hidden]
  http:127.0.0.1:8080
Hit CTRL-C to stop the server
```

This commit ensures that the slashes will be there. It doesn't effect the actual work, it's just representational mistake. Some people, including me, may want to copy and paste the URL from there.


gokaygurcan:
meh. no need anymore.

PR #199, opened 2015-10-24
proxy-pass option to avoid changing Host header

developit:
This adds a `proxy-pass` boolean option, which prevents the proxy from changing the origin of the host header to the target URL.

See `changeOrigin` option here: https://github.com/nodejitsu/node-http-proxy#options

### DIFF
--- a/bin/http-server
+++ b/bin/http-server
@@ -9,6 +9,7 @@ var colors     = require('colors'),
     opener     = require('opener'),
     argv       = require('optimist')
       .boolean('cors')
+      .boolean('proxy-pass')
       .argv;
 
 var ifaces = os.networkInterfaces();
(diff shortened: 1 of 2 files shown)


0xR:
This is actually quite a old PR. I need this functionality as well. Did you find a workaround by now? @developit

angelkenneth:
I had to copy and paste the `bin/http-server` and `lib/http-server.js` files to my own project just to i can set `changeOrigin: true`.

Still hoping this will be merged.

fabioimpe:
@indexzero Please I this this functionality too, is there any chance too see this PR merged?

Thank you

thornjad:
This looks reasonably straightforward and useful! However there are some merge conflicts which need to be resolved before this can move forward, @developit could you take a look at those?

thornjad:
Given the variety of options available in http-proxy, this may be better implemented with the approach of passing through options to http-proxy itself in #459, what do you think?

thornjad:
Closing in favor of #495

PR #202, opened 2015-10-28
Pointing to scripts online is harmful

dabockster:
The installation portion to the readme instructs the user to use a shell script provided by npm to install the package manager. This is harmful because it leaves open way too many ambiguities. A safer way would be to instruct the user to use a package manager to install npm (even though most Node installers provide npm by default now).

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,11 +6,7 @@
 
 # Installing globally:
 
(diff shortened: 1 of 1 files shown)


indexzero:
Thanks. That's ancient documentation.

PR #194, opened 2015-10-14
Push state option

eerwitt:
First try at implementing an option to support push state by rewriting all request URLs to use `/index.html` instead of the URL requested.

Does this conflict with other options? I'm concerned about changing the requested URL.

This was originally requested as part of  indexzero/http-server#80

Issue #80, opened 2014-07-05
Support HTML5 push-state


mmahalwy:
How can I use angularjs' html5Mode true. 

Need to do some server configuration. What needs to be done? Where?

HipsterZipster:
+1

lmadeira:
Same here. +1

astoilkov:
+1

bumblebeeman:
If you are using ASP.Net MVC I wrote a couple of articles which cover it, I imagine other frameworks are similar in concept at least - http://www.codeproject.com/Articles/806029/Getting-started-with-AngularJS-and-ASP-NET-MVC-Par

maxdumas:
+1

super-coder:
@bumblebeeman you tutorial is overkill, all you have to do is routes.MapRoute(name: "Default",url: "{*anything}",defaults: new{controller = "Home",action = "Index",}); and then return your index.html in Index method of Home controller.
But how to do it without ASP.NET MVC and where to host?

DarkReZuS:
+1

luisrudge:
still no solution to this?

super-coder:
what do you mean, it is resolved, use the snippet above

luisrudge:
I'm not using aspnet mvc, so it's not resolved :)

super-coder:
i watched a course by PluralSight "Building AngularJS and Node.js Apps with the MEAN Stack" they do it with Node.js in few first lessons. Also have a look here http://stackoverflow.com/questions/17777967/using-angularjs-html5mode-with-nodejs-and-express

super-coder:
you don't have to call Node REST API then, you can just use Node to help Angular work in html5Mode=true

luisrudge:
You're missing the point. I want to use this project, `http-server`, and be able to handle pushState urls.

katmai7:
+1

pensierinmusica:
+1

derhuerst:
> I used superstatic to solve the problem: https://www.npmjs.com/package/superstatic

Why not use this?

herrherrmann:
+1

zwacky:
+1

if you tackle this, please try to make it usable without a config file, like superstatic curenlty does it.

indexzero:
A PR is welcome for this if it can be done in an efficient manner.

barroudjo:
[...]

PR #205, opened 2015-11-03
Created fallback option and handling

riceboyler:
For those of us using http-server to serve up SPAs (like a simple React site, for instance), we need to be able to provide a fallback address to return to if the route isn't found, so that it can be handled by the SPA framework's router.

This creates a fallback option and provides a test to ensure it works as designed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,4 +72,6 @@ This will install `http-server` globally so that it may be run from the command 
 
 `-r` or `--robots` Provide a /robots.txt (whose content defaults to 'User-agent: *\nDisallow: /')
 
+`-f` or `--fallback` Provide a fallback url if response returns 404 - useful for SPA frameworks
+
 `-h` or `--help` Print this list and exit.
--- a/bin/http-server
+++ b/bin/http-server
@@ -38,6 +38,7 @@ if (argv.h || argv.help) {
     '  -K --key     Path to ssl key file (default: key.pem).',
     '',
     '  -r --robots  Respond to /robots.txt [User-agent: *\\nDisallow: /]',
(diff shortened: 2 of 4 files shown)


vlkosinov:
+1

justincy:
+1 Exactly what I wanted. I'll be using this fork for now.

justincy:
NM, that's not exactly what I wanted. This is implemented as a redirect. I don't see how this would work very well for SPAs. What I want is more like proxy/url rewriting. In other words, when a page isn't found, I want the index page to be served but the url to be kept the same.

justincy:
I guess https://github.com/indexzero/http-server/issues/80 is what I want.

indexzero:
This looks like a reasonable feature, but will wait for a few more +1s before merging to make sure that it is really something folks in the community of users want.

princed:
Here you go :+1:

huston007:
+1

maxmaximov:
+1

indexzero:
I think that #194 is a better option for this.

mikehedman:
+1

tyrsius:
+1

zhanjiaxia:
+1

numical:
+1

dhdanno:
I noticed when using fallback, the original query string does not seem to be passed to index.html

foxel:
+1

IceReaper:
+1 Need this feature to redirect assets/ to a shared ../assets/ directory.

riceboyler:
Closing this as #194 is a more elegant solution

PR #206, opened 2015-11-05
Fix URI protocol output on start (`//` added)

smileart:


smileart:
@indexzero This one fixes helpful output added in #141 which appears to be broken now in `master`.

gokaygurcan:
Then, you'll have `http:////` because of line 145.

smileart:
👋

PR #208, opened 2015-11-09
Report correct server address

glls:
When you specify an address with -a, http-server reports that it listens to all the available interfaces but serves only the address specified (as it should)


indexzero:
Thank you for your contribution!

PR #218, opened 2015-12-04
Adds --hide-dotfiles option

johntron:
Hooks up [ecstatic](https://github.com/jfhbrook/node-ecstatic)'s `showDotfiles` option - allows users to hide files beginning with a dot (`.`). Required updating ecstatic to 1.3.1.

I've also written a test, but wasn't able to run existing tests, so haven't committed.

### DIFF
--- a/bin/http-server
+++ b/bin/http-server
@@ -38,6 +38,7 @@ if (argv.h || argv.help) {
     '  -K --key     Path to ssl key file (default: key.pem).',
     '',
     '  -r --robots  Respond to /robots.txt [User-agent: *\\nDisallow: /]',
+    '  --no-dotfiles  Do not show dotfiles',
     '  -h --help    Print this list and exit.'
   ].join('\n'));
   process.exit();
@@ -99,7 +100,8 @@ function listen(port) {
     robots: argv.r || argv.robots,
     ext: argv.e || argv.ext,
     logFn: logger.request,
-    proxy: proxy
+    proxy: proxy,
+    showDotfiles: argv.dotfiles
   };
 
   if (argv.cors) {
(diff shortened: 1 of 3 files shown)


.vimrc
```diff
@@ -0,0 +1,3 @@
+set ts=2
+set sw=2
+set expandtab
```
indexzero: Please do not include your personal dotfiles in this PR.

bin/http-server
```diff
@@ -38,6 +38,7 @@ if (argv.h || argv.help) {
     '  -K --key     Path to ssl key file (default: key.pem).',
     '',
     '  -r --robots  Respond to /robots.txt [User-agent: *\\nDisallow: /]',
+    '  --hide-dotfiles  Don\'t show dotfiles',
```
indexzero: This should be `--no-dotfiles` to file conventions for CLI apps.

indexzero:
Thank you for your contribution. I had a few minor comments. This seems like a reasonable feature, but I am going to wait to merge it until there are +1s from the community of `http-server` users to prove it's usefulness in the aggregate.

derhuerst:
:+1: for this

bin/http-server
johntron: Fixed - I didn't know optimist had that feature. Does http-server just follow optimist's approach, or are the conventions outlined somewhere else?

johusman:
:+1: I want to keep .git-folders in the structure without exposing them. Surprised this feature wasn't available.

BigBlueHat:
Rebased and merged in https://github.com/indexzero/http-server/commit/402ad49faea69ad19fc9cca408446612907760b7

Which GitHub seems to have missed...

Thanks for adding this (way back when) @johntron!

PR #240, opened 2016-01-26
allow to use log file, and allow clean logfile on the fly

ybootin:
Just a little feature to specify an output log file (instead of using http-server > logfile.txt),
plus add a feature to clear the log file on the fly by calling `/clear-log` .

### DIFF
--- a/bin/http-server
+++ b/bin/http-server
@@ -7,6 +7,8 @@ var colors     = require('colors'),
     httpServer = require('../lib/http-server'),
     portfinder = require('portfinder'),
     opener     = require('opener'),
+    util       = require('util'),
+    fs         = require('fs'),
     argv       = require('optimist')
       .boolean('cors')
       .argv;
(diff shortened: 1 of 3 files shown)


bin/http-server
```diff
@@ -48,11 +53,39 @@ var port = argv.p || parseInt(process.env.PORT, 10),
     ssl = !!argv.S || !!argv.ssl,
     proxy = argv.P || argv.proxy,
     utc = argv.U || argv.utc,
+    logfile = (function () {
+      var file = argv.logfile || argv.l;
+      if (!file) {
+        return '';
+      }
```
thornjad: Could we check for these options before jumping into this function, thus avoiding requiring evaluation at all?

bin/http-server
```diff
@@ -48,11 +53,39 @@ var port = argv.p || parseInt(process.env.PORT, 10),
     ssl = !!argv.S || !!argv.ssl,
     proxy = argv.P || argv.proxy,
     utc = argv.U || argv.utc,
+    logfile = (function () {
```
thornjad: Instead of a messy IIFE, could we put this checking into its own function, and call it here?

thornjad:
Looks like there are some merge conflicts to take a look at as well

thornjad:
Closing due to inactivity. For now, a good way to capture logging is regular shell redirection

PR #241, opened 2016-01-29
Made it possible to handle errors when using proxy. To avoid exception.

ducklord:
If the proxy forwards to a server that returns a 404 for that file, this will make the proxy throw an exception, and in return also crash http-server.

### DIFF
--- a/lib/http-server.js
+++ b/lib/http-server.js
@@ -109,6 +109,9 @@ function HttpServer(options) {
         changeOrigin: true
       });
     });
+    if (options.proxyErrorHandler) {
+      proxy.on('error', options.proxyErrorHandler);
+    }
   }
 
   var serverOptions = {


yalishizhude:
Why don't merge this branch?

mik01aj:
+1. It's just 3 lines, come on! :)

lib/http-server.js
```diff
@@ -109,6 +109,9 @@ function HttpServer(options) {
         changeOrigin: true
       });
     });
+    if (options.proxyErrorHandler) {
+      proxy.on('error', options.proxyErrorHandler);
```
thornjad: How would `proxyErrorHandler` be specified?

thornjad:
@ducklord are you still interested in moving forward with this PR?

ducklord:
Not important to me, any more. We could not use this software because of this, so we changed to something different many years ago.

PR #242, opened 2016-02-03
feat(cors): Adds support for Credentials header

andreibaragan:
Hello,

I've enabled the CORS Credentials support in corser. In a nutshell, if Access-Control-Allow-Credentials is enabled then Access-Control-Allow-Origin should have the value of the Origin header and not '*'.


thornjad:
Closing in favor of #434

PR #247, opened 2016-02-15
Fixes #246: Add option --auto to set proxy option autoRewrite

### LINKED ISSUE #246: Add option --auto to set proxy option autoRewrite
Enhancement to allow setting of --auto flag on command line to pass to proxy, allowing auto rewrite behavior to be turned on.

kwesterfeld:
- add handling for --auto flag

Fixes #246

### DIFF
--- a/bin/http-server
+++ b/bin/http-server
@@ -9,6 +9,7 @@ var colors     = require('colors'),
     opener     = require('opener'),
     argv       = require('optimist')
       .boolean('cors')
+      .boolean('auto')
       .argv;
 
 var ifaces = os.networkInterfaces();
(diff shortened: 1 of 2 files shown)


bin/http-server
```diff
@@ -32,6 +33,7 @@ if (argv.h || argv.help) {
     '  -U --utc     Use UTC time format in log messages.',
     '',
     '  -P --proxy   Fallback proxy if the request cannot be resolved. e.g.: http://someurl.com',
+    '  --auto       Set proxy autoRewrite flag for proxy, handling 301/302 responses',
```
thornjad: Instead of changing `auto` to `autoRewrite` later, I think it would be simpler and more clear to use the option `--auto-rewrite` from the start instead. Otherwise I think this looks good.

thornjad:
@kwesterfeld looks like we have some merge conflicts to take a look at

thornjad:
Given the variety of options available in http-proxy, I'm going to close this in favor of the approach of passing through options to http-proxy itself in #459, which would solve this same issue

Issue #215, opened 2015-12-03
EADDRINUSE instead of finding a free port


thorn0:
As per #18, if the default port 8080 is in use, some other free port should automatically get found and used. However, this doesn't happen (Node 5.1, Windows 7 x64).

```
d:\foo>http-server
events.js:141
      throw er; // Unhandled 'error' event
      ^

Error: listen EADDRINUSE 0.0.0.0:8080
    at Object.exports._errnoException (util.js:856:11)
    at exports._exceptionWithHostPort (util.js:879:20)
    at Server._listen2 (net.js:1234:14)
    at listen (net.js:1270:10)
    at net.js:1379:9
    at doNTCallback3 (node.js:463:9)
    at process._tickCallback (node.js:369:17)
    at Function.Module.runMain (module.js:459:11)
    at startup (node.js:138:18)
    at node.js:974:3
```

kmgilbert100:
Works on node 0.10.41 ... Not working on node 4.2.2

kmgilbert100:
Havn't checked any versions in between but looks like it can't find a port on 4.2.2 or greater

lmx99:
you can refer to that url
http://superuser.com/questions/262942/whats-different-between-ctrlz-and-ctrlc-in-unix-command-line/262948#262948

BigBlueHat:
This is fixed when using the default port via the `portfinder` upgrade in 989fa1c7f5727da92ac727f673a2d4b7a121124a. It does *not* handle fallbacks when using the `-p` option. There's a pending/possible option for that which is under discussion in #363.

PR #261, opened 2016-03-23
update portfinder version to fix auto increment port

w3cj:
Old version of portfinder is broken and does not find an open port.

Fixes issue: https://github.com/indexzero/http-server/issues/215

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "http-proxy": "^1.8.1",
     "opener": "~1.4.0",
     "optimist": "0.6.x",
-    "portfinder": "0.4.x",
+    "portfinder": "~1.0.2",
     "union": "~0.4.3"
   },
   "devDependencies": {


derhuerst:
@w3cj 

> `"portfinder": "~1.0.2"`

What about using a [caret (`^`) range](https://github.com/npm/node-semver#caret-ranges-123-025-004).

w3cj:
Currently the exact version (0.4.0) of `portfinder` is set in the [package.json](https://github.com/indexzero/http-server/blob/master/package.json#L67) and `portfinder` [recently went from 0.4.0 to 1.x](https://github.com/indexzero/node-portfinder/releases)

I'm for ~ range to at least get patches, but avoid any compatibility issues with minor versions.

derhuerst:
> I'm for ~ range to at least get patches, but avoid any compatibility issues with minor versions.

I'd assume that `portfinder` adheres to semver and therefore usually doesn't have compatibility problems with minor bumps. The value of semver stands or falls by it's adoption.

cmckni3:
I agree with @derhuerst on this. Use and adopt semver but do some sort of QA.

Personally, I prefer `1.x` over `^1.0.2` but these are equivalent.

BigBlueHat:
989fa1c7f5727da92ac727f673a2d4b7a121124a sets the `portfinder` dependency to `^1.0.13`--which I hope does the trick for everyone! If now, PRs welcome. 😄

PR #272, opened 2016-05-06
Update readme.md (Add Live demo/tutorial)

theredstapler:


derhuerst:
Nice idea!

Just a minor remark: The tutorial can be made even simpler :wink:  `npm init` is not necessary and `./node_modules/.bin/http-server` is the most straightforward way to run the server.

BigBlueHat:
Great video @theredstapler, but I think we'll pass on adding it to the README for the reasons mentioned by @derhuerst and that if you've found the README, you're likely copying and pasting out of it vs. watching a video. Great "getting started" content, though!

PR #277, opened 2016-05-24
Fix #276 add command line option --contentType

### LINKED ISSUE #276: Add --content-type command line options
Hello,

Thanks for this very useful package !

I'd like to be able to do 

```
 http-server [path] --contentType "custom/type"
```

that will use the node `contentType` options and set the default value of server's content type.

SebastienLAURET:
see #276

### DIFF
--- a/bin/http-server
+++ b/bin/http-server
@@ -104,7 +104,7 @@ function listen(port) {
   };
 
   if (argv.contentType) {
-     options.contentType = argv.contentType;
+    options.contentType = argv.contentType;
   }
 
   if (argv.cors) {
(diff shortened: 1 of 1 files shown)


bin/http-server
```diff
@@ -24,6 +24,7 @@ if (argv.h || argv.help) {
     '  -i           Display autoIndex [true]',
     '  -e --ext     Default file extension if none supplied [none]',
     '  -s --silent  Suppress log messages from output',
+    '  --contentType Define the content-type default value',
```
thornjad: I know this is a tiny nitpick, but could you align "Define the..." with "Enable CORS..." below?

thornjad (changes requested):
I know this is an old PR, but all it needs is documentation added to the README

zbynek:
With #693 it will be possible to set content type for each file extension separately, so maybe this option is redundant (?)

PR #279, opened 2016-05-24
Fixes #278: adds support for verbose access logging

### LINKED ISSUE #278: Add --verbose to log out the headers of each request
I'd like to be able to see the exact headers that were requested by a given http request.

This functionality would be useful for debugging or inspecting unknown client requests.

Not sure if `--verbose` is the best name for such a flag, but its the best I can think of right now.

scottopell:
Fixes #278 

Use `-v` or `--verbose` to enable verbose access logging.

Looks like this:

```
scratch/http-server - [master] » node bin/http-server -p 8000 -v
Starting up http-server, serving ./public
Available on:
  http://127.0.0.1:8000
  http://192.168.0.2:8000
Hit CTRL-C to stop the server
[Tue May 24 2016 12:54:02 GMT-0400 (EDT)] "GET /" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/50.0.2661.102 Safari/537.36"
HEADER                    VALUE
host                      localhost:8000
connection                keep-alive
cache-control             max-age=0
accept                    text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8
upgrade-insecure-requests 1
user-agent                Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/50.0.2661.102 Safari/537.36
accept-encoding           gzip, deflate, sdch
accept-language           en-US,en;q=0.8
if-none-match             "2609389-210-"2016-05-22T20:20:38.000Z""
if-modified-since         Sun, 22 May 2016 20:20:38 GMT
```

Adds dependency on [`columnify`](https://www.npmjs.com/package/columnify#columnify-objects) for pretty printing the headers when enabled.

### DIFF
--- a/bin/http-server
+++ b/bin/http-server
@@ -7,6 +7,7 @@ var colors     = require('colors'),
     httpServer = require('../lib/http-server'),
     portfinder = require('portfinder'),
     opener     = require('opener'),
+    columnify  = require('columnify'),
     argv       = require('optimist')
       .boolean('cors')
       .argv;
@@ -38,6 +39,7 @@ if (argv.h || argv.help) {
     '  -K --key     Path to ssl key file (default: key.pem).',
     '',
     '  -r --robots  Respond to /robots.txt [User-agent: *\\nDisallow: /]',
+    '  -v --verbose Log out the full headers of each request along with the typical content',
     '  -h --help    Print this list and exit.'
   ].join('\n'));
   process.exit();
@@ -68,6 +70,14 @@ if (!argv.s && !argv.silent) {
           date, req.method.cyan, req.url.cyan,
           req.headers['user-agent']
         );
+        if (argv.v || argv.verbose) {
+          var headerString = columnify(req.headers,
+                                       { columns: ['Header', 'Value'] });
+          logger.info(
+            '%s',
+            headerString
+          );
+        }
       }
     }
   };
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
   ],
   "dependencies": {
     "colors": "1.0.3",
+    "columnify": "^1.5.4",
     "corser": "~2.0.0",
     "ecstatic": "^1.4.0",
     "http-proxy": "^1.8.1",


mwcz:
Yes please!

vitalets:
Would love to get this merged!

vitalets:
Hi @indexzero @BigBlueHat @thornjad !
What do you think about merging this pr? It is really very useful feature!

bin/http-server
```diff
@@ -7,6 +7,7 @@ var colors     = require('colors'),
     httpServer = require('../lib/http-server'),
     portfinder = require('portfinder'),
     opener     = require('opener'),
+    columnify  = require('columnify'),
```
thornjad: I would prefer that we either don't add a new dependency, or use it for all logging (and the help output).

bin/http-server
```diff
@@ -68,6 +70,14 @@ if (!argv.s && !argv.silent) {
           date, req.method.cyan, req.url.cyan,
           req.headers['user-agent']
         );
+        if (argv.v || argv.verbose) {
```
thornjad: #131 will merge soon, be prepared to change how these are accessed

thornjad: Looking back at this, #131 might take a while due to extenuating circumstances, so don't worry about it!

bin/http-server
```diff
@@ -38,6 +39,7 @@ if (argv.h || argv.help) {
     '  -K --key     Path to ssl key file (default: key.pem).',
     '',
     '  -r --robots  Respond to /robots.txt [User-agent: *\\nDisallow: /]',
+    '  -v --verbose Log out the full headers of each request along with the typical content',
```
thornjad: We now have `-v` implemented as the short option for `--version`, which I think is fine. This option could be just `--verbose` with no short version.

On the other hand, since http-server is pretty verbose by default, I wonder if this might be better called `--very-verbose`?

bin/http-server
thornjad: I am not a big fan of adding this dependency. It does make the headers look better, but I'm not sure we need the extra baggage to do that. I think `util.format` should do just fine without any extra libraries

bin/http-server
thornjad: Could we also move this up next to `--silent` to keep it all together?

thornjad: This also should to be added to the README

thornjad:
Closing due to inactivity, please re-open if we'd like to revisit this. I'll leave the associated issue open since it would be nice to have this functionality

PR #288, opened 2016-06-22
Adding contentType cli option to address #276

howardroark:


Jameskmonger:
Hey @howardroark, couple of linting issues on this failed the CI build.

howardroark:
Thanks @Jameskmonger ! Fixed ;)

borisirota:
@Jameskmonger Hi, will this be merged ?

It is very useful in my case where I have static website and prefer the urls to be of a form `/about` without the extension `/about.html`.
The files are saved in the file system without the extension but need to be served with the correct `Content-Type` in order to be parsed correctly.

Just a small note, its not very correct that `Use a custom Content-Type response header for all requests`.
As I see in the [ecstatic source code](https://github.com/jfhbrook/node-ecstatic/blob/master/lib/ecstatic.js#L196), the --contentType param is being used only if the `mime.lookup` fails. IMO its great because it means that this param is a fallback rather than override of expected behavior (e.g. override the `Content-Type` header of a javascript file to html file's content type).

mmahut:
Oh boy.

PR #292, opened 2016-07-13
Fix "/public" directory so files can be accessed

NikhilPeri:
From /lib to access /public the correct directory should be ../public

### DIFF
--- a/lib/http-server.js
+++ b/lib/http-server.js
@@ -33,11 +33,11 @@ function HttpServer(options) {
   }
   else {
     try {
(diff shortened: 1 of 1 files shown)


derhuerst:
AFAIK, `./public` is relative to the current working directory of the process, *not* to directory `http-server.js` resides in. Therefore, `./public` should be correct.

Consider this:

```shell
mkdir public
touch public/foo
git clone https://github.com/indexzero/http-server.git http-server
node http-server/bin/http-server
```

PR #293, opened 2016-07-14
Add ecstatic's gzip mode

### LINKED ISSUE #56: gzip support
Would it be possible to have an option for gzip/deflate compression?

SteveVanOpstal:
closes #56


gaearon:
Very cool. We started recommending this package in https://github.com/facebookincubator/create-react-app for testing prod builds so it would be great to see gzipped bundle sizes. 😄

gaearon:
Ah, never mind, I misunderstood this change.

PR #309, opened 2016-09-17
Fix crash on Windows Subsystem for Linux

kevin2r:
Since "Windows Subsystem for Linux" hasn't yet implemented the Linux API to list the network interfaces, it was crashing when running `os.networkInterfaces();` method.

### DIFF
--- a/bin/http-server
+++ b/bin/http-server
@@ -127,7 +127,7 @@ function listen(port) {
       '\nAvailable on:'.yellow
     ].join(''));
 
-    if (argv.a || host !== '0.0.0.0') {
+    if (host === '0.0.0.0') {
       logger.info(('  ' + protocol + canonicalHost + ':' + port.toString()).green);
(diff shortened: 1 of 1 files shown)


derhuerst:
2985845 is definitely a breaking change an should be discussed adequately.

kevin2r:
With the change in the "if" statement I'm looking for a way to not execute the `os.networkInterfaces();` when we don't need to, by default (with no parameters) **'http-server'** should listen on all interfaces, ie **127.0.0.1** as the documentation describes **-a Address to use (defaults to 0.0.0.0)**. The rest of the code behaves correctly, is only that log section (which I'm changing) the part that behaves a little confusing for me.

thornjad:
Knowing that a lot has improved in WSL since this PR was opened, I'm going to close for now. Please reopen if the issue is found to persist!

PR #311, opened 2016-09-19
Use safe colors to avoid header problems

lnwdr:
The `union` module iterates over a map of headers. If the supplied `headers` object is a thing and `colors` is loaded in it's "normal" (not "safe") mode, `union` will try to set headers like "bold" and "red" which contain shell color codes. Node will complain about these invalid header character and throw an exception:

```
_http_outgoing.js:351
      throw new TypeError('The header content contains invalid characters');
      ^

TypeError: The header content contains invalid characters
    at ServerResponse.OutgoingMessage.setHeader (_http_outgoing.js:351:13)
    at ResponseStream.writeHead (/Users/lnwdr/temp/testproject/node_modules/http-server/node_modules/union/lib/response-stream.js:78:21)
```

### DIFF
--- a/bin/http-server
+++ b/bin/http-server
@@ -2,7 +2,7 @@
 
 'use strict';
 
-var colors     = require('colors'),
+var colors     = require('colors/safe'),
     os         = require('os'),
     httpServer = require('../lib/http-server'),
     portfinder = require('portfinder'),
--- a/bin/http-server
+++ b/bin/http-server
@@ -59,14 +59,14 @@ if (!argv.s && !argv.silent) {
       if (error) {
         logger.info(
           '[%s] "%s %s" Error (%s): "%s"',
-          date, req.method.red, req.url.red,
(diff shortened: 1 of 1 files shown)


indexzero:
+1, but as implemented this will cause a number of errors. You need to change all strings using `colors` currently to use the `colors/safe` code style of invocation. e.g. [this code](https://github.com/indexzero/http-server/blob/master/bin/http-server#L67-L71) would become:

``` js
logger.info(
  '[%s] "%s %s" "%s"',
  date, colors.cyan(req.method), colors.cyan(req.url),
  req.headers['user-agent']
);
```

lnwdr:
Ok, changed it accordingly. Tests are passing and I tried it. Color output works.

lnwdr:
What about this issue? Seems ready to merge.

BigBlueHat:
Thanks for this @lnwdr.

PR #320, opened 2016-10-12
Add .npmignore

hmil:
Files which are not useful to the end user, such as tests and build resources should be excluded from the npm package as it saves everyone bandwidth and time.

I'm not sure if the `public` folder is used only for tests or if it is useful to run http-server. If not, please feel free to add it or let me know and I'll add it for you.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,3 @@
+test
+screenshots
+.travis.yml


indexzero:
I believe we also want what is currently in `.gitignore`. Please add that to this file.

hmil:
Sure.
Why is there `!node_modules/node-static` in .gitignore?

derhuerst:
I'd strongly advise to use a `files` field in `package.json`, which acts as a *whitelist* (`.npmignore` acts as a *blacklist*). This prevents unnecessary stuff ending up in the npm bundle in a more robust way.

hmil:
@derhuerst That's a good point, I updated the PR.

BigBlueHat:
Thanks for this addition @hmil!

PR #323, opened 2016-10-28
Added cli option --pid that creates pidfile

Zhomart:


bin/http-server
```diff
@@ -111,6 +113,19 @@ function listen(port) {
     }
   }
 
+  if (argv.pid) {
+    fs.writeFileSync(argv.pid, process.pid ,{ flag: 'w+' });
+    process.on('exit', function () {
+      try {
+        fs.unlinkSync(argv.pid);
```
derhuerst: In Node.js, synchronous operations should never be executed during runtime, as they block the event loop. Use `fs.unlink`.

Zhomart:
@derhuerst You're right. I discovered forever only after this PR. I'll close it.

PR #324, opened 2016-10-28
corsHeader was changed to header, fixing XSS

KevinEverywhere:
When CORS was not working, I noticed that the http-server-test.js used "header" instead of "corsHeader" for the option sent to the server. This was a quick fix for my use, but by defining the entire object instead of adding to it, this will need to be revisited for those sending additional header information to the server. Once this is done, you can use CORS like this:

http-server --cors='Access-Control-Allow-Origin':'*','Access-Control-Allow-Credentials':'true'

### DIFF
--- a/bin/http-server
+++ b/bin/http-server
@@ -107,7 +107,7 @@ function listen(port) {
   if (argv.cors) {
     options.cors = true;
     if (typeof argv.cors === 'string') {
-      options.corsHeaders = argv.cors;
+      options.headers = argv.cors;
     }
   }
 


BigBlueHat:
@KevinEverywhere it looks like this change would effectively overwrite any other `headers` sent in.

`ecstatic` now supports (and possibly did earlier too...) it's own bunch of `cors` defaults:
https://github.com/jfhbrook/node-ecstatic/blob/fe17a5623c18319f82101b6549e9e2e526e03d1f/lib/ecstatic/opts.js#L91-L97

It then allows `headers` to overwrite those or modify them.

In `http-server`'s case, we've got this additional `corsHeaders` system:
https://github.com/indexzero/http-server/blob/master/lib/http-server.js#L68-L78

This has the advantage of just allowing the `Access-Control-Allow-Headers` header to be appended to--without re-declaring that entire line in your config/parameters.

What I'd like to be sure of is that...
 - `options.cors` turns on the default "enable CORS" headers (as they do in `ecstatic`)
 - `options.corsHeaders` appends to the permitted headers list seen in `Access-Control-Allow-Headers`
 - `options.headers` can be used to add headers to all requests

If I've misunderstood something you were working to fix here, please let me know!

Thanks!
🎩

thornjad:
Closing due to inactivity and the issues already discussed. Please re-open if these changes should be revisted.

Issue #325, opened 2016-11-02
EADDRINUSE 0.0.0.0:8086 error for multiple instances


shershen08:
Recently started having following error when trying to start more than one instance of whs.
```
Error: listen EADDRINUSE 0.0.0.0:8086
    at Object.exports._errnoException (util.js:1036:11)
...
```
whs version - 3.10.9
node version v6.6.0

console output looks like this

```
$ whs -p 4444
Websocket Server Listening on Port: 8086
Starting up http-server, serving ./ on: http://127.0.0.1:4444
Hit CTRL-C to stop the server
events.js:160
      throw er; // Unhandled 'error' event
```

BigBlueHat:
As of `0.10.0` if you've *not* specified a specific port via the `-p` option, then one will be auto-selected for you by `portfinder` (now upgraded to `1.0.13`.

That said, the example you've listed specifies port (via `-p 4444`), so that would still throw that error.

However, we're discussing a `--find-port` option in #363 and your input there would be welcome!

thornjad:
Comments are welcome in #363, but for now I'm going to close this since a) the latest versions of http-server should find a new port by default, and b) it is intended behavior that trying to run multiple instances on the same port is an error.

PR #332, opened 2016-12-06
replace portfinder with newport

harttle:
Fixes: #325 and #215

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "http-proxy": "^1.8.1",
     "opener": "~1.4.0",
     "optimist": "0.6.x",
-    "portfinder": "0.4.x",
+    "portfinder": "^1.0.13",
     "union": "~0.4.3"
   },
   "devDependencies": {


derhuerst:
By using `newport`, `http-server` loses the ability to specify a specific port (which is VERY handy imo) and fallback to the next available. Also, the defacto standard port for adhoc/dev HTTP servers is `8080`.

harttle:
I think it's faster and also valid to let Node.js server to use a random clean port :)

derhuerst:
Still, this is a regression. As some people use this for serving static files permanently, `http-server` will  effectively be broken for them.

xuexb:
This is one of my ideas : Default port is `8080`, if the error is random.

```js
process.on('uncaughtException', function (err) {
  if (err.syscall === 'listen' && err.port === 8080) {
    console.log(colors.red('\n8080 port is occupied, will randomly generate a new port!\n'));
    newport(function (err, port) {
      if (err) { throw err; }
      listen(port);
    });
  }
});

if (!port) {
  portfinder.basePort = 8080;
  portfinder.getPort(function (err, port) {
    if (err) { throw err; }
    listen(port);
  });
}
else {
  listen(port);
}
```
image

Thanks~

derhuerst:
> This is one of my ideas : Default port is `8080`, if the error is random.

> If set up over the port, when being given an error I think it is reasonable. If no port is set, the next port is randomly enabled when the `8080` is occupied for the service to start

I'd personally rather like the opposite: I would like to specify a port (and have it fail if the port is used). In case I didn't care about a specific port and just want *any* port, I would pass `--find-port`, which uses `newport`.

This would work out using *only* `newport`, without catching via `uncaughtException`, and without being a regression over the previous behavior.

Af few notes on your implementation:

- Using `process.on('uncaughtException', cb)` is considered dangerous & bad practice. Note how `process.once` would be more appropriate.
- `console.log(colors.red(` should be replaced by something that writes to `stderr` (instead of `stdout`) and sets the exit code to non-`0`.
- Your code uses `portfinder` & `newport`. Why not use one of them?

xuexb:
Above is just one of my ideas, really lack of thinking~  Thank you very much~

> I'd personally rather like the opposite: I would like to specify a port (and have it fail if the port is used). In case I didn't care about a specific port and just want any port, I would pass `--find-port`, which uses `newport`.

This is good~

BigBlueHat:
I'm assisting @indexzero in catching up with the backlog of PRs here.

@derhuerst I'm 👍 on the approach you've outlined, and would be happy to accept a PR.

@harttle would you be in support of the approach outlined in https://github.com/indexzero/http-server/pull/332#issuecomment-268971162

I'm not keen to loose the ability to define a specific port, but having the *option* to fallback / have one found for me, sounds fab!

harttle:
@BigBlueHat OK, I'll make a PR this week later.

harttle:
I've implemented the `--find-port` option. Please ping this thread if there's still anything I can help.

BigBlueHat:
@harttle just took this for a spin locally (bash in Windows 10) and it worked like a charm!

I'm going to give this a bit for others to test, then possibly add it to the v0.10.0 release.

Much thanks!
🎩

harttle:
I just find out that upgrade the `portfinder` fixes this issue without any regression. The last commit would be better.

BigBlueHat:
@harttle oh. Even better. Mind squashing this PR into just the necessary code?

harttle:
It's Ok, there is only one commit actually, since I reset it before.☺️

BigBlueHat:
@harttle looks like this is *just* the `portfinder` update, correct? If that's the case, that means this code doesn't currently provide the `--find-port` option. Would you mind adding that to this PR?

Thanks!

BigBlueHat:
@harttle I'm going to close this one--as it's a bit tangled now--in favor of the newly opened #363 which is specifically about the `--find-port` param.

I've upgraded `portfinder` in  989fa1c by way of keeping things up to date.

Thanks for taking the lead on this feature!
🎩

PR #350, opened 2017-03-04
Added compression middleware under --compression to gzip responses

radum:
Creating those .gz files could add to much work for dev. This is why if we compress the response by default using a cli flag might be better.


radum:
This will actually break the stream on Chrome Canary 😢 
I'm closing it for now.

PR #358, opened 2017-04-18
set HTTP response header for gzip responses

cvan:


yamboy1:
If you put the `else {` from line 63 onto line 64 it will fix part of the build error. The other thing to fix is to get rid of that template literal. Travis is testing this on v4.1 and v0.12. 0.12 doesn't support template literals.
This is what I  would change the template literal to:
`'gzip, ' + this.headers['Content-Encoding']`

yamboy1:
You can add that template literal now. Support has been dropped for version 0.10.

cvan:
@yamboy1 thanks, updated

yamboy1:
Add this to the start of your file: `/*jshint esversion: 6 */`. It should fix the build error.

cvan:
done; tests passing 😄

BigBlueHat:
@cvan can you bring me up to speed on why this is needed?

From what I can tell this will make the statement that all resources are gzip'd even if a `.gz` file is not found.

`ecstatic` already sets the header properly if/when a compressed version of the file exists:
https://github.com/jfhbrook/node-ecstatic/blob/fe17a5623c18319f82101b6549e9e2e526e03d1f/lib/ecstatic.js#L179-L212

Let me know what wasn't working for you!

yangdonglai:
@BigBlueHat 
For example,I have many *.pbf file  what are actual  .gzip file. 
they created by tool and used for  [mapbox-gl](https://www.mapbox.com/mapbox-gl-js/example/simple-map/) ( a visual lib for map).

BigBlueHat:
@yangdonglai can you describe this situation more completely? What do the files look like on disk, and what headers are missing/being-sent? Thanks!

yangdonglai:
@BigBlueHat below is the right response , you can notice the *content-encoding*=*gzip*  and *content-type*=*application/x-protobuf*. 
image

below is error response when I use hs to server static pbf file,  there is not a  *content-encoding* header
image
which will cause below error when use [mapbox-gl](https://www.mapbox.com/mapbox-gl-js/example/simple-map/)

image

BigBlueHat:
Thanks for this @yangdonglai. This actually looks primarily like a `Content-Type` related issue. Protobuf hasn't yet [registered a media type with IANA](https://www.iana.org/assignments/media-types/media-types.xml), so the [mime-types](https://github.com/jshttp/mime-types) library doesn't (apparently) have record of `pbf` files.

However, your first screenshot shows an express server properly serving the file with that `application/x-protobuf` media type. Express also uses the `meda-types` library, so I'm wondering if this express server used for the first example has some other settings applied.

The gzip thing seems to also be an issue (on the current release), but the primary problem here seems to be the media type detection thing.

BigBlueHat:
@cvan could you add some tests specific to this patch? Thanks!

thornjad:
@cvan wouldn't this cause all requests to have a gzip header? With this PR, if gzip is enabled, it would send the gzip header with anything, such as `index.html`, which is not gzipped, right? That could cause browsers to try to decode non-encoded content.

thornjad:
@cvan do you have interest in continuing with this pull request?

cvan:
> @cvan do you have interest in continuing with this pull request?

I don't currently have the bandwidth, my apologies. thanks for checking :+1:

thornjad:
No problem! If you don't mind, I'm going to close this for now. Please do reopen if you want to get back to it, and/or anyone else can take a look if they want!

Issue #363, opened 2017-05-01
Add `--find-port` parameter


BigBlueHat:
#332 explored a `--find-port` parameter that was promising for situations when one wants to fallback from the default or defined port if it's unavailable...which will solve #325 & #215 bothersomeness.

harttle:
989fa1c upgrades `portfinder`, which will solve #325 & #215. The new version of [`portfinder`][portfinder] will guarantee  the `port` found to be a free port.

In the current codebase, [`portfinder.basePort = 8080`][line84] sets the default port to `8080`, `portfinder` will find a free port (i.e. 8081, 8082, and so on.) if `8080` is not available.

[line84]: https://github.com/indexzero/http-server/blob/master/bin/http-server#L84
[portfinder]: https://github.com/indexzero/node-portfinder

BigBlueHat:
So, the current `0.10.0` release does handle the fallback for the default port, but does not provide a fallback port--and still throws the error mentioned in #215 & #325.

Adding this `--find-port` option could fix that, but I'm not quite sure it's a good idea as specifying a port likely means the developer/operator expects that port to be available and if it is not, there *should* be an error...rather than just another port selected.

Open to input, though!

ahmetb:
I think this should be implemented as `-p 0`  (port zero) flag . It's the unix convention for many server applications that is telling the OS to assign a random port. See #386.

thornjad:
I'm inclined to agree with @ahmetb that `-p 0` seems the more "usual" way of specifying this. And, while http-server doesn't support _random_ port usage with `-p 0`, it should already work to walk upward from 8080 until it finds an open port. @BigBlueHat feel free to re-open this if you disagree, but I'm closing for now since `-p 0` already supports this functionality.

thornjad:
This functionality is not currently documented, I'm adding that now.

PR #364, opened 2017-05-03
Feature: add `--find-port` parameter

harttle:
Find a free port when the port given by `-p` is not available. Solves #363

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,8 @@ This will install `http-server` globally so that it may be run from the command 
 
 `-U` or `--utc` Use UTC time format in log messages.
 
+`-F` or `--find-port` Find a free port when the port given by `-p` is not available.
+
 `-P` or `--proxy` Proxies all requests which can't be resolved locally to the given url. e.g.: -P http://someurl.com
 
 `-S` or `--ssl` Enable https.
--- a/bin/http-server
+++ b/bin/http-server
@@ -31,6 +31,7 @@ if (argv.h || argv.help) {
     '  -c           Cache time (max-age) in seconds [3600], e.g. -c10 for 10 seconds.',
     '               To disable caching, use -c-1.',
     '  -U --utc     Use UTC time format in log messages.',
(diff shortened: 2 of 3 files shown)


ahmetb:
I think this should be implemented as `-p 0` flag. See #386.

BigBlueHat:
@harttle would you have time/interest in writing some tests for this new feature? The code looks good, and the use case you expressed in https://github.com/indexzero/http-server/issues/386#issuecomment-320376241 may not be that uncommon (i.e. wanting a port "above" a certain number). Thanks!

harttle:
This should be sufficient, yet not complete. Since `bin/http-server` wasn't tested before, it's hard to do complete test for it. For instance, the stdout/stderr.

thornjad:
Given the discussion in #386, would this feature maybe be better implemented as a `--base-port` which is the same as the default behavior of no `--port` given, but searching begins at `--base-port`?

My concern with the `-p XXXX --find-port` pattern is that the user is both specifying a port _and_ expecting that the port may end up something other than what is specified. I think it may be more clear if an error is thrown when the port the user wants is in use. Then if the user wants to search for a port, they _don't_ use `-p` and may optionally determine the base port to start at.

@BigBlueHat @harttle @ahmetb any thoughts on a `--base-port` approach?

sumitparakh:
What is the status of this PR?

harttle:
Sorry I forgot this for a white and deleted that forked repo unexpectedly, I tried and could not restore that commit which is needed to continue patching this pull request. It's confirmed by [this question from stackexchange](https://softwareengineering.stackexchange.com/questions/61577/should-i-keep-my-github-forked-repositories-around-forever). If that's the way, we should open another PR for #386 and this should be closed. @thornjad

sumitparakh:
Hi @harttle @thornjad . Would you mind if i try this?

thornjad:
Closing since the fork is gone, and since we have #544

Issue #360, opened 2017-04-25
Add `.headers` support


BigBlueHat:
The `.headers` file idea comes from the [File Handler in wptserve](http://wptserve.readthedocs.io/en/latest/handlers.html#file-handlers) (a Python-based server for running the Web Platform Tests system at the W3C).

They look like this:
```
Content-Type: text/html
```

In `wptserve` they can be associated with single files (`index.html.headers` or `index.headers`) as well as directories (for `/about/` style requests) via a `__dir__.headers` file (because Python-esque).

The advantage is that you can then provide any number of HTTP headers for any file or "directory" or perhaps even non-resource requests such as handling HEAD-only requests or `303 See Also` specific entities.

This could be used to address a number of issues related mentioned here related to caching and content-types as well as things like advanced CORS and CSP handling.

Thoughts welcome!

nbarbettini:
I'd love to see this merged into http-server. It's harder than I expected to find a simple Node HTTP server that lets me set custom headers easily.

LvChengbin:
I am sorry, I cannot understand how to use it. I added a `.headers` file under the dir, but it doesn't work.

BigBlueHat:
@LvChengbin it's not implemented yet, but there seems to be enough interest in it to do the work now.

@nbarbettini thanks for the +1. I'd been a bit concerned that this changed the nature of http-server a bit, but given that these are *optional* files that augment requests, I think it's fine--though I'll still probably put it behind an optional command line flag to avoid the double-disk checking for each request if it's not in use.

BigBlueHat:
Netlify uses a top-level `_headers` file with path mapping:
https://www.netlify.com/docs/headers-and-basic-auth/

Having the header files closer to the actual requested file seems better, but a single top-level file does avoid the double-file disk reads per each request (so that's a good).

aggregate1166877:
Out of desperation I've forked all this and merged hedefalk's 7 year old version with the latest stuff [here](https://github.com/aggregate1166877/http-server).

It's a temporary hack solution and I'm likely not going to merge future changes into that, so it's officially unmaintained, but it's a workaround for the desperate like me until the authors merge in hedefalk's changes.

PR #365, opened 2017-05-13
Added possibility to use reroutes

SiXoS:
Added possibility to use reroutes. It works by specifying a JSON file with the -R parameter. The json file is a list which is looped through for each http request. It does a find-replace call on the url of the request with the defined reroute rules. If the url is changed, a log is printed (except if you are in silent mode). I have added the following simple reroute to the example in bublic:

```
[
   {
     "match": {"regexp":"/img/([a-z]+).png", "flags":"i"},
     "replace": "/img/$1_reroute.png"
   }
]
```

This will reroute all png images to /img/\<original name\>_reroute.png.

You can also chain the reroute by specifying `abortOnMatch: false`. For example, lets say that we want to move the reroute images to a different folder instead without changing the existing rule:

 ```
[
   {
     "match": {"regexp":"/img/([a-z]+).png", "flags":"i"},
     "replace": "/img/$1_reroute.png",
     "abortOnMatch": false
   },
   {
      "match": {"regexp":"/img/([a-z]+)_reroute.png", "flags":"i"},
      "replace": "/img/rerouted/$1.png"
   }
]
```


SiXoS:
I'll fix the stylechecks immediately!

brandonros:
Nice work man! What do you personally use this for?

SiXoS:
I have a small typescript library which is a SPA. I'm using History to navigate. When I refresh the page I get 404 as that page doesn't exist. So I'm basically rerouting all non-static paths to index.html.

BigBlueHat:
Rewriting URLs is great and potentially quite useful. However, I think #369 is probably a better (because simpler) for solving this History hacking scenario.

Additionally, while I think the URL rewriting could be quite useful for other reasons, I'd like to avoid creating a new JSON format for dealing with redirect URL encoding.

Lastly, I think a clearer use case (or two) for this would also be fabulous. If the default doc solves the History hacking situation, then we'll need another reason for possibly adding URL rewriting.

There's also some URL rewriting history laying about here that's probably worth a dig through--as well as ecstatic issues and such.

Thanks for proposing this idea! Let me know if #369 would fix your particular situation.

BigBlueHat:
Oh. And tests. Of course. :grin:

SiXoS:
Yes, #369 would solve my problem and is a more straightforward solution.

You can also handle all failing requests with the fallback feature so I guess that is a better solution.

BigBlueHat:
Thanks again for submitting this, @SiXoS. I'm going to close this for now (at least) until we get a few more things sorted out with some of the similar conversations going on. I'm not opposed to a rerouting system, but would like it to be as standards (or prevailing pattern) based as possible. Appreciate you starting this conversation, though!

BigBlueHat:
I'd actually like to revisit the concept of a URL rewriter for this project. However, I'd like it to use some common rewriter format found in OpenAPI, CouchDB, or similar. It doesn't have to bee too painfully powerful, but some basic to/from handling could go along way--including solving for #369 issues without directly the Web. 😸

SiXoS:
What about using the [Firebase way](https://firebase.google.com/docs/hosting/url-redirects-rewrites)? I would imagine a lot of people that uses this tool uses Firebase hosting and only uses it locally for development. It is very simple though. They could then use the same rules for the server and locally. Of course, this is just an assumption and it is very likely that I'm way off track.

BigBlueHat:
@SiXoS nice find! It's quite similar to the [CouchDB's rewrites](http://docs.couchdb.org/en/2.0.0/api/ddoc/rewrites.html#rewrite-section-is-an-array), but I like the inclusion of headers, redirects, and rewrites in one place.

There's also a prevailing pattern in the [OpenAPI](http://openapis.org/) spec (and similar) to use [URI Templates](https://tools.ietf.org/html/rfc6570) for defining "changeable" parts of a URI. I built a little playground for that eons ago, if you want to take a look: http://bigbluehat.github.io/vue-uri/

There's still a concern I'm working through about introducing a config file. The tagline here is "a simple zero-configuration command-line http server." This is a completely additional "config" similar to my [`.headers` proposal](https://github.com/indexzero/http-server/issues/360)...so...it becomes a "meta" question about the future of the project, not just the technical feasibility or even usefulness of it.

However, I'm keen to have something like this available (for myself and others), so let's discuss it more! :grin:

I'll open an issue for that and mention this PR.

Would you be open to converting this code to using the JSON format from Firebase (at least as a starting point)?

Issue #367, opened 2017-05-13
Is there a way to redirect all requests to index.html to support client side routing?


dumptyd:

dumptyd:
The -P flag, yes.

dumptyd:
But the server crashes when if I try to access 127.0.0.1/something (when `something` is handled by client side)

stack trace:

```
http-server -P /
Starting up http-server, serving ./
Available on:
  http://127.0.0.1:8080
  http://192.168.1.115:8080
Unhandled requests will be served from: /
Hit CTRL-C to stop the server
[Sat May 13 2017] "GET /boo" "Mozilla/5.0 (X11; Linux x86_64; rv:54.0) Gecko/20100101 Firefox/54.0"
/usr/lib/node_modules/http-server/node_modules/requires-port/index.js:13
  protocol = protocol.split(':')[0];
                     ^

TypeError: Cannot read property 'split' of null
    at required (/usr/lib/node_modules/http-server/node_modules/requires-port/index.js:13:22)
    at Object.common.setupOutgoing (/usr/lib/node_modules/http-server/node_modules/http-proxy/lib/http-proxy/common.js:102:7)
    at Array.stream (/usr/lib/node_modules/http-server/node_modules/http-proxy/lib/http-proxy/passes/web-incoming.js:120:14)
    at ProxyServer.<anonymous> (/usr/lib/node_modules/http-server/node_modules/http-proxy/lib/http-proxy/index.js:80:21)
    at Array.<anonymous> (/usr/lib/node_modules/http-server/lib/http-server.js:109:13)
    at dispatch (/usr/lib/node_modules/http-server/node_modules/union/lib/routing-stream.js:119:21)
    at g (events.js:292:16)
    at emitNone (events.js:86:13)
    at emit (events.js:185:7)
    at /usr/lib/node_modules/http-server/node_modules/union/lib/routing-stream.js:114:23
```

BigBlueHat:
@dumptyd as you've found, `-P` is not what you want. That's for proxying other endpoints and `/` isn't one.

Would you mind giving PR #369 a try for your use case and reporting thoughts over there?

Thanks!
🎩

BigBlueHat:
We've added the recommendation to use `404.html` as the fallback URL. It prevents over catching (at multiple URLs) of a single "fallback" response, still provides the option to later create static (or server-side rendered) resources at each URL (which is *far* better for production environments), and still plays nice with `pushState()` usage.

Issue #338, opened 2017-01-20
How to add catch all redirect?


Altiano:
I need something like this

`app.get("*", callback)`

so every request url will be redirected to index.html. 
but since this package not using express, I don't know how to solve this.

derhuerst:
See #30, #306, #271 and #123. They all seem to be bundled in #80.

Altiano:
gonna read it then, thanks a lot :)

Altiano:
I'm not sure, #80 doesn't provide any clear answer, 

I think all I need to do is just modifying some lines here: https://github.com/indexzero/http-server/blob/master/lib/http-server.js

But turns out, it probably more complex than that?

derhuerst:
> I'm not sure, #80 doesn't provide any clear answer,

No, it is the proposal for this feature.

houston3:
-P http://localhost:8080/? worked for me. Note the question mark at the end.

thornjad:
The `-P http://localhost:8080?` method is rather brilliant, and I think that may be the answer here, since your `index.html` is essentially acting as the proxy for all other pages.

BigBlueHat:
@thornjad let's get this added to the README.

Sending `200 OK` is an anti-HTTP pattern. While SPA's may "benefit" from it, it's bad architecture both generally and for deployed SPAs and should really only be used in development and not production. Even the Angular docs suggest it shouldn't be used permanently because it's bad for both UX and performance.

However, having it handy for development does help folks who are doing HTML and not HTTP, so solving it at the README layer sounds great. :smiley:

vovarudomanenko:
if you are getting an error "zsh: no matches found", escape the question mark like shown in example below
```
npx http-server --proxy http://localhost:8080/\?
```
[more detailed explanation](https://www.reddit.com/r/zsh/comments/h9mdvc/why_do_i_get_this_error_zsh_no_matches_found/)

Issue #331, opened 2016-12-05
with  react-router error  404


qq645381995:
with react-router  browserHistory  http  404    

redirect  index.html ?

derhuerst:
See #80.

Issue #318, opened 2016-10-11
Add fallback to index.html


kirill-konshin:
Please add fallback to index.html for every "not real" file, pretty much the same way as `connect-history-api-fallback` does (https://github.com/bripkens/connect-history-api-fallback#introduction):
1. The request is a GET request
2. which accepts text/html,
3. is not a direct file request, i.e. the requested path does not contain a . (DOT) character and
4. does not match provided pattern

Not sure if I am not creating a duplicate, but I haven't found one... I only found a fork: https://github.com/riceboyler/http-server.

derhuerst:
There is files without `.` in the name. Why do you want to show a fallback?

kirill-konshin:
Fallback is needed when you have a single-page app and you would like to route all 404s to the app first and let it figure out whether it has a matching route or not.

It is very rare that anyone would server files without a dot, usually those paths are handled dynamically by the app, they are not real files served from FS.

derhuerst:
> Second, a comfortable way to serve single page applications. This might become a possible duplicate of spa-server, spaserve and devserve. Issues regarding this: #69 and #95, #7, #80, #254.

– https://github.com/indexzero/http-server/issues/63#issuecomment-193244575

ashelley:
@kirill-konshin found your issue, not sure if this helps but if you read here:

https://github.com/indexzero/http-server/pull/194

you can npm-install -g spa-http-server and you will get a --push-state option

kirill-konshin:
Create React App is using `pushstate-server`, btw.

darlanmendonca:
@ashelley @indexzero I tried option --push-state with modules `spa-http-server`. and works.

but one thing that I dont understand, the modules `spa-http-server` on npm, target to repo `http-server`, and have a differente readme, with mention to --push-state.

I'm confuse, why this feature dont exists on original `http-server`

derhuerst:
@darlanmendonca Probably it has been published from a fork.

kud:
For people who need it:

```
$ npm install spa-http-server --save-dev
```

```
"test:dev": "npm run build && http-server dist -p 8080 --push-state",
```

Note that `spa-http-server` replaces `http-server` bin.

kud:
However I wonder why this option is not in this current project.

BigBlueHat:
It's not in this project because it's not good HTTP. Serving identical representations to any request with no ability to throw errors except "fake ones" shown to the user in JavaScript cuts against the grain of the Web.

[...]

PR #369, opened 2017-05-20
Fallback option for single page applications

skleeschulte:
I implemented the option to specify a default url path that is used to respond to unresolved requests (e.g. client requests /login which does not exist on the server -> server responds as if asked for /index.html). This is very helpful when serving a single page app, where everything that does not resolve to a resource on the server shall be served with the SPA entry file (e.g. index.html) *without redirecting* the client.

I chose to implement a "default url" rather than a "default file path", because this way it is straight forward to use the existing ecstatic instance: if the ecstatic middleware does not handle the initial request, `req.url` is simply replaced with the configured default url and the request is passed to the same ecstatic middleware again.

If somebody wants to use this before it is (or if it is not) merged:
`npm install skleeschulte/http-server --save`
And then use `-f` or `--fallback` parameter:
`http-server ./root -f /index.html`

Related issues: #7 #80 #198 #286 #318 #331 #338 #367

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,8 @@ This will install `http-server` globally so that it may be run from the command 
 
 `-P` or `--proxy` Proxies all requests which can't be resolved locally to the given url. e.g.: -P http://someurl.com
 
+`-D` or `--default` Default url path to serve if the request cannot be resolved. e.g.: /index.html
+
 `-S` or `--ssl` Enable https.
 
 `-C` or `--cert` Path to ssl cert file (default: cert.pem).
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ This will install `http-server` globally so that it may be run from the command 
 
 `-P` or `--proxy` Proxies all requests which can't be resolved locally to the given url. e.g.: -P http://someurl.com
 
-`-f` or `--fallback` Default url path to serve if the request cannot be resolved. e.g.: /index.html (Fallback will only be used for GET requests for .htm and .html files.)
+`-f` or `--fallback` Default url path to serve if the request cannot be resolved. e.g.: /index.html (Fallback will only be used for GET requests with an URL ending with `/`, `.htm` or `.html`.)
 
 `-S` or `--ssl` Enable https.
 
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ This will install `http-server` globally so that it may be run from the command 
 
 `-P` or `--proxy` Proxies all requests which can't be resolved locally to the given url. e.g.: -P http://someurl.com
 
-`-D` or `--default` Default url path to serve if the request cannot be resolved. e.g.: /index.html
+`-f` or `--fallback` Default url path to serve if the request cannot be resolved. e.g.: /index.html
 
 `-S` or `--ssl` Enable https.
 
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ This will install `http-server` globally so that it may be run from the command 
 
 `-P` or `--proxy` Proxies all requests which can't be resolved locally to the given url. e.g.: -P http://someurl.com
 
-`-f` or `--fallback` Default url path to serve if the request cannot be resolved. e.g.: /index.html (Fallback will only be used for GET requests with an URL ending with `/`, `.htm` or `.html`.)
+`-f` or `--fallback` Default url path to serve if the request cannot be resolved. e.g.: /index.html (Fallback will only be used for GET requests with an URL ending with a path segment without an extension or with one of the extensions `.htm` or `.html`.)
 
 `-S` or `--ssl` Enable https.
 
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ This will install `http-server` globally so that it may be run from the command 
 
 `-P` or `--proxy` Proxies all requests which can't be resolved locally to the given url. e.g.: -P http://someurl.com
 
-`-f` or `--fallback` Default url path to serve if the request cannot be resolved. e.g.: /index.html
+`-f` or `--fallback` Default url path to serve if the request cannot be resolved. e.g.: /index.html (Fallback will only be used for GET requests for .htm and .html files.)
 
 `-S` or `--ssl` Enable https.
 
--- a/bin/http-server
+++ b/bin/http-server
@@ -18,29 +18,29 @@ if (argv.h || argv.help) {
     'usage: http-server [path] [options]',
     '',
     'options:',
-    '  -p            Port to use [8080]',
-    '  -a            Address to use [0.0.0.0]',
-    '  -d            Show directory listings [true]',
-    '  -i            Display autoIndex [true]',
-    '  -g --gzip     Serve gzip files when possible [false]',
-    '  -e --ext      Default file extension if none supplied [none]',
-    '  -s --silent   Suppress log messages from output',
+    '  -p             Port to use [8080]',
+    '  -a             Address to use [0.0.0.0]',
+    '  -d             Show directory listings [true]',
+    '  -i             Display autoIndex [true]',
+    '  -g --gzip      Serve gzip files when possible [false]',
+    '  -e --ext       Default file extension if none supplied [none]',
+    '  -s --silent    Suppress log messages from output',
     '  --cors[=headers]   Enable CORS via the "Access-Control-Allow-Origin" header',
     '                     Optionally provide CORS headers list separated by commas',
-    '  -o [path]     Open browser window after starting the server',
-    '  -c            Cache time (max-age) in seconds [3600], e.g. -c10 for 10 seconds.',
-    '                To disable caching, use -c-1.',
-    '  -U --utc      Use UTC time format in log messages.',
+    '  -o [path]      Open browser window after starting the server',
+    '  -c             Cache time (max-age) in seconds [3600], e.g. -c10 for 10 seconds.',
+    '                 To disable caching, use -c-1.',
+    '  -U --utc       Use UTC time format in log messages.',
     '',
-    '  -P --proxy    Fallback proxy if the request cannot be resolved. e.g.: http://someurl.com',
-    '  -D --default  Default url path to serve if the request cannot be resolved. e.g.: /index.html',
+    '  -P --proxy     Fallback proxy if the request cannot be resolved. e.g.: http://someurl.com',
+    '  -f --fallback  Default url path to serve if the request cannot be resolved. e.g.: /index.html',
     '',
-    '  -S --ssl      Enable https.',
-    '  -C --cert     Path to ssl cert file (default: cert.pem).',
-    '  -K --key      Path to ssl key file (default: key.pem).',
+    '  -S --ssl       Enable https.',
+    '  -C --cert      Path to ssl cert file (default: cert.pem).',
+    '  -K --key       Path to ssl key file (default: key.pem).',
     '',
-    '  -r --robots   Respond to /robots.txt [User-agent: *\\nDisallow: /]',
-    '  -h --help     Print this list and exit.'
+    '  -r --robots    Respond to /robots.txt [User-agent: *\\nDisallow: /]',
+    '  -h --help      Print this list and exit.'
   ].join('\n'));
   process.exit();
 }
@@ -49,7 +49,6 @@ var port = argv.p || parseInt(process.env.PORT, 10),
     host = argv.a || '0.0.0.0',
     ssl = !!argv.S || !!argv.ssl,
     proxy = argv.P || argv.proxy,
-    defaultUrlPath = argv.D || argv.default,
     utc = argv.U || argv.utc,
     logger;
 
@@ -104,7 +103,7 @@ function listen(port) {
     ext: argv.e || argv.ext,
     logFn: logger.request,
     proxy: proxy,
-    defaultUrlPath: defaultUrlPath
+    fallback: argv.f || argv.fallback
   };
 
   if (argv.cors) {
--- a/bin/http-server
+++ b/bin/http-server
@@ -34,7 +34,7 @@ if (argv.h || argv.help) {
     '',
     '  -P --proxy     Fallback proxy if the request cannot be resolved. e.g.: http://someurl.com',
     '  -f --fallback  Default url path to serve if the request cannot be resolved. e.g.: /index.html',
-    '                 (Fallback will only be used for GET requests with an URL ending with /, .htm or .html.)',
+    '                 (Fallback will only be used for GET requests with an URL ending with a path segment without an extension or with one of the extensions .htm or .html.)',
     '',
     '  -S --ssl       Enable https.',
     '  -C --cert      Path to ssl cert file (default: cert.pem).',
--- a/bin/http-server
+++ b/bin/http-server
@@ -34,6 +34,7 @@ if (argv.h || argv.help) {
     '',
     '  -P --proxy     Fallback proxy if the request cannot be resolved. e.g.: http://someurl.com',
     '  -f --fallback  Default url path to serve if the request cannot be resolved. e.g.: /index.html',
+    '                 (Fallback will only be used for GET requests for .htm and .html files.)',
     '',
     '  -S --ssl       Enable https.',
     '  -C --cert      Path to ssl cert file (default: cert.pem).',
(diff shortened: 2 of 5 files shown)


skleeschulte:
BTW: If you prefer something other than `-D` and `--default` for the command line options, I'm happy to change that. I have seen `-f` and `--fallback` - would that be better?

BigBlueHat:
@skleeschulte the `--fallback` name is indeed better than `--default`. It's essentially going to route all 404's to this file--which, in many cases could be a `404.html` I'd reckon.

And...actually...now that I've typed that... `ecstatic` already provides a default/fallback file in the form of `404.html`:
https://github.com/jfhbrook/node-ecstatic/blob/fe17a5623c18319f82101b6549e9e2e526e03d1f/lib/ecstatic.js#L128

What's concerning me now is that this change (or something like it) would effectively eliminate the value of `404` errors--making all requests end in a `200`...which at the very least make debugging things hard, but also seems unwise from a web architecture perspective.

Early, you mentioned...
> This is very helpful when serving a single page app, where everything that does not resolve to a resource on the server shall be served with the SPA entry file (e.g. index.html) without redirecting the client.

Is this because you're doing "history hacking" as in #365?

Thanks for any more info you can provide!

skleeschulte:
@BigBlueHat The ecstatic 404 fallback returns the status code 404, whereas for a SPA entry file it should be 200.

And yes, this is for doing "history hacking", or in other words, for utilizing the browser's history API for navigation within a single page application. When doing this while not using server side rendering, it is common to return the application entry point (e.g. index.html) for every GET request that does not point to an actual resource on the server. It is then the single page application's task to show an error message (e.g. "404"), if the path is not valid (because the server cannot know if a path is valid, except if every valid path is in some way configured on the server, which can be difficult to keep in sync with the app).

BigBlueHat:
@skleeschulte can you point me to some resources where this is being encouraged for SPA's? This feels like an anti-pattern (though I know I've seen it done many a time).

If you're doing that sort of "client-side only routing," I'd recommend *not* doing "history hacking" and instead sticking to the client-side only routing space in the fragment identifier space (i.e. `#/about`).

It just feels terribly un-webby... Send me some links, though, and I'll do my own digging. Thanks!

skleeschulte:
@BigBlueHat Isn't it enough proof that there is a requirement for this if you have seen it done many times? ;)

One source: The [react-router](https://github.com/ReactTraining/react-router) package uses the [history](https://github.com/ReactTraining/history) package, which offers `createBrowserHistory` (HTML5 history API, whithout #) and `createHashHistory` (for "legacy browsers", with #). In the [react-router docs](https://reacttraining.com/react-router/web/guides/quick-start), almost all examples use the `BrowserRouter`, [which uses the "browser history"](https://github.com/ReactTraining/react-router/blob/d601722f8515eea3910bb3a69313f3f2fc37df16/packages/react-router-dom/modules/BrowserRouter.js#L3) (without #). They also [reccomended to not use the "hash history" in production](https://stackoverflow.com/a/35185338/846987) at one time, although I cannot find the original source anymore.

gregpasha:
I also want this feature.[ React Router v4 recommends against using it's hash history (except for legacy browser support) ](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/api/HashRouter.md)
 and favoring browser history. Adding this feature means I can just deploy a static file web server such as nginx rather than deploying node and configuring it to send all of my routes back to index.html. Thank you so much for considering putting this in.

BigBlueHat:
> @BigBlueHat Isn't it enough proof that there is a requirement for this if you have seen it done many times? ;)

Yeah...that's never a good reason to do anything. :grin:

I do understand the use case presented here, however, I still feel it's an anti-pattern (perhaps a pervasive one). Additionally, the Progressive Web Apps world encourages the opposite--at least in relation to full screen apps: https://developers.google.com/web/fundamentals/native-hardware/fullscreen/#keep_the_user_in_the_fullscreen_experience

It's essentially like symlink-ing any file request to return the results of another file...which may (or may not!) do anything meaningful with the request. It also means the "fallback" file will have to "spoof" a 404 request that only the person behind the keyboard understands--as there's no way (to my knowledge) for JS to tell the browser that the URL in the location bar is *actually* 404, 410, 302, etc...

I think there's a clearer hope for the future in the world of ServiceWorkers and "catching" client-side requests--while still maintaining some mapping to things on the server.

However, I'll ponder this a bit more and let others weigh in. 😃 Especially if folks can provide non-React specific issues, that'd be super. Thanks!

mafredri:
I would also like to see support for this. It would allow me to easily test SPAs locally that would otherwise sit behind Nginx and behave the same way. For SPA's I think it's relatively uncommon to use the # fragment for navigation these days, unless you're supporting IE9 (even then it's usually fallback behavior).

Take Angular for example, the default behavior of the angular-router is to use `history.pushState` to change the URL, it does support hash-based URLs, but that is not the default.

Regarding the spoofing of 404's I'd argue that it's irrelevant. The consumers of SPAs are users and given that the SPA can display the error message to the user, there is no problem to solve. Would it make sense to do `curl` against an SPA and expect a meaningful HTTP response code? I think not. There are edge cases, of course, say a file is removed and now `curl` returns 200 with HTML instead of the expected file, but then again, I don't think this option would be used in production anyway.

mafredri:
I might add that the Angular docs do have a section on this behavior and why HTML5 style URLs are preferred under [Appendix: LocationStrategy and browser URL styles](https://angular.io/docs/ts/latest/guide/router.html#!#browser-url-styles).

Essentially it boils down to this:

> Almost all Angular projects should use the default HTML5 style. It produces URLs that are easier for users to understand. And it preserves the option to do server-side rendering later.

> Rendering critical pages on the server is a technique that can greatly improve perceived responsiveness when the app first loads. An app that would otherwise take ten or more seconds to start could be rendered on the server and delivered to the user's device in less than a second.

> This option is only available if application URLs look like normal web URLs without hashes (#) in the middle.

Gcastelo:
This is a great feature and the timing is great as we're looking for this to support Angular 2 routes. I've tested it and it works. When do you think it'll be merged to master?

BigBlueHat:
Still chewing on this one. I'm curious if [devserve](https://github.com/hoho/devserve) might work for the simple SPA use case. It has fallback to single HTML file and proxying. Would that work for most of you? Or is there a key thing `http-server` offers that you want in addition to this requested feature?

mafredri:
Biggest advantage IMO would be to be able to use one server for most of your use-cases. While `devserve` only relies on node internals, I'm a bit skeptical of using anything that hasn't been maintained in two years. `http-server` also has more options, e.g. serving gzipped files, which makes it more attractive.

Finally, and this is very much a personal opinion for which I hope you do not take offense, not supporting this (extremely widespread) use-case leaves very little reason for `http-server` to remain installed. I would rather have one server installed that I can use for all of my quick use-cases.

Gcastelo:
I agree with what @mafredri says. It's a common feature in web servers. Not supporting leaves a bit of a hole. 

For what it's worth, we actually decided to move to nginx as we only need to serve an angular 2 bundle, but are more comfortable to use it in the long run due to the set of features, if needed, and the available alpine docker image makes it a third of the size of an alpine node+http-server image.

petermunnings:
One of the problems is that webpack dev server gives you that option:

```
  devServer: {
    historyApiFallback: true
  }
```
Once you build for production and run it on http-server your site routes suddenly stop working.

So one of the most common dev tools leads you along a development path that is not supported in production.  By the time you are there you don't want to go back and redesign, you just want a quick fix that does what your dev webserver was doing.

Not saying its right or wrong, just saying that is how it is and most devs who have gone down that route (myself included) would at least like the option of switching on a fallback path

SiXoS:
I think this would be a good idea as well. I don't see any real problem with "fake" 404 pages. It makes it easier to create error pages that fits the site. Otherwise you'll have to include all the resources in the error page files as well.

Regarding the hard-to-debug part, logging all the fallbacks would make it easy to quickly discover a page that shouldn't trigger a fallback.

skleeschulte:
As suggested by @SiXoS, I added logging to the fallback function. In addition, I limited the fallback function to only change requests that meet the following conditions:

- The request is a GET request.
- The request URL ends with `/`, `.htm` or `.html`.

Not sure if the last condition covers all use cases. It prevents serving the fallback file for requests for e.g. favicon.ico, robots.txt and sitemap.xml.

BigBlueHat:
@skleeschulte check the `Accept` headers. Never sniff URLs to discover their content types--they're opaque (or should be treated as such). Proof: https://github.com/BigBlueHat/this-is-not-a.pdf 😁 

Also, after much research, I'm more in favor of a general URL rewriter--which could be misused for this purpose if you really must...--but would serve a much wider use case that doesn't run counter to the awesomeness that is HTTP. 😉 

Here's a great video on some performance comparison of SPA's vs. *uncached* server-rendered resources: https://www.youtube.com/watch?v=J2dOTKBoTL4&feature=youtu.be&t=8m58s

I'm not keen to directly encourage the use of this 200 fallback approach (however popular) as it smells like an anti-pattern that fixes some tedium (the `#` in URLs) with a nuke (breaking HTTP, URLs, caches, browser back/forward optimizations, JS processing pipelines, bookmarks, archival tools, etc, etc).

I'd be more than happy to consider a general purpose URL rewriter option/config-system, fwiw--and consequently will reopen https://github.com/indexzero/http-server/pull/365 shortly.

Alternatively, there are many SPA-focused dev servers worth checking out instead:
* https://www.npmjs.com/package/superstatic
* https://www.npmjs.com/package/spa-http-server
* https://github.com/scottcorgan/pushstate-server

BigBlueHat:
Just came across this project which pre-renders defined routes as static pages: https://github.com/graphcool/prep

Might be worth a look for far faster SPA sites. Cheers.

xfg:
This function must be in http-server.

BigBlueHat:
@xfg there are many other servers/packages that include this feature ([as mentioned](https://github.com/indexzero/http-server/pull/369#issuecomment-307281380)).

Is there a particular reason you feel *this* package "must"?

vegaskev:
FWIW, I had great success with pushstate-server and running `pushstate-server dist/ 4200 > ./output.log &` to fire up the server package and run it in the background.  'dist/' being my dist folder for my angular app and 4200 being my port.  Yours may be different.  My config is ubuntu with node in the background and a nginx reverse proxy serving up my dist folder under ssl.

I totally prefer http-server, but as mentioned by multiple people http-server does not support html5 routing.  I did notice a npm package spa-http-server, which I tried and failed.  I also tried the popular superstatic server with no luck.  pushstate-server was the only one that I was able to achieve html5 routing.

Also, IMO, I would love to go back to http-server.  Nothing against pushstate-server, I'm just more familiar with http-server package  :)  ⭐️ ⭐️ ⭐️ ⭐️ ⭐️  support from what I can see.

Hope this helps somebody.

xfg:
@BigBlueHat yes there are other servers but this server is more popular among remaining and because similar servers basically are used for a rapid deployment of development/testing environment but not for  production so this feature is a 'must have' for this server. Currently the developers can't use this server for a development SPA and forced to install and configure more heavy web servers like nginx or use less popular and respectively more vulnerable to errors web servers about you are mentioned. This all awfully uncomfortable.

jackjwilliams:
I would like to see support for this as well. I was sad to see it does not exist for http-server, but glad there is discussion around it.

The web, she's a changing. I see no reason for the option not to be added, for use in SPAs specifically. 

Either way, thank you for the http-server package

BigBlueHat:
@jackjwilliams mainly, it's not what this project is about. It's an HTTP server. Limiting it's response to just 200 is [406](https://http.cat/406). 😸 

There are an ample number of "fallback" servers that are built with the express (heh) purpose of providing this feature for folks that want to build SPAs.

For folks that want to use HTTP, there's `http-server`.

Given the growing amount of research that advises *against* client-side rendering (CSR) (see https://github.com/indexzero/http-server/pull/369#issuecomment-307281380), your users will be happiest if you prerender (see https://github.com/indexzero/http-server/pull/369#issuecomment-308778195)--as stated in the quote in this comment https://github.com/indexzero/http-server/pull/369#issuecomment-304735562.

Don't use `pushState` unless you're returning some *unique* content at that URL. If you've no plans to do that (ever), use hash URLs.

In the meantime (or if you disagree), any of these SPA focused servers will let you do what you need:

* https://www.npmjs.com/package/superstatic
* https://www.npmjs.com/package/spa-http-server
* https://github.com/scottcorgan/pushstate-server

This project, however, will remain focused on serving HTTP.

Thanks,
🎩

Tim15:
... and now I have to find another server to use

namnm:
Looking forward to seeing this reopen!

jackjwilliams:
I don't think that will happen - @BigBlueHat has stated that it is not this repos intention to support that.

ReeganExE:
I'm using `npm install -g serve` as suggested from `create-react-app`.

joe-oli:
http-server serves up a 404.html when a route / page is not found; so why not duplicate your index.html file and rename it 404.html.. works for me, am i missing something?

nayfin:
I can confirm. Dropping 404.html into dist folder of an Angular app provides requested functionality.

BigBlueHat:
Thanks, @nayfin! I'll get this added to the README shortly. Cheers!

Devnco:
Following the idea of @joe-oli   it could be great to have a new flag to specify the file that will be retrieved as 404 page, we could precise the index.html as placeholder.

BigBlueHat:
@Devnco that would probably be best done as a change to `node-ecstatic` (which is the core of `http-server`) instead:
https://github.com/jfhbrook/node-ecstatic/blob/1b5692c5ae3af10018d9c1d4a9ff3a1a7ee6be01/lib/ecstatic.js#L348-L355

Devnco:
@BigBlueHat Indeed that surely the best place 👍

csbenjamin:
> This project, however, will remain focused on serving HTTP

@BigBlueHat Serving index.html when requesting another path seams like "serving HTTP" for me. Am I missing something?

nickfaughey:
So `$ cp index.html 404.html` is seriously the recommended approach? Having this line in my Dockerfile makes me gag:

```
RUN npm run build && \
    cp dist/index.html dist/404.html
```

 I'll stick with `http-server` for now just to not have to deal with the friction of migrating an existing project, but there's little reason to continue to use this package in the future if it can't serve up SPA's cleanly.

BigBlueHat:
@nickfaughey fwiw, you can also use `http-server -P http://localhost:8080?` (note the trailing `?`) to proxy your `index.html` file. That should get you what you want without the file copy. We'll be adding that to the README soon. Cheers!

RossThomson:
> @nickfaughey fwiw, you can also use `http-server -P http://localhost:8080?` (note the trailing `?`) to proxy your `index.html` file. That should get you what you want without the file copy. We'll be adding that to the README soon. Cheers!

This is great unless it's HTTPS/self-signed certificates, where you get "[0] Error: self signed certificate"

wangpengfeido:
> @nickfaughey fwiw, you can also use `http-server -P http://localhost:8080?` (note the trailing `?`) to proxy your `index.html` file. That should get you what you want without the file copy. We'll be adding that to the README soon. Cheers!

It is perfect

wangpengfeido:
> @nickfaughey fwiw, you can also use `http-server -P http://localhost:8080?` (note the trailing `?`) to proxy your `index.html` file. That should get you what you want without the file copy. We'll be adding that to the README soon. Cheers!

It is perfect

Undistraction:
> In the meantime (or if you disagree), any of these SPA focused servers will let you do what you need:
> 
> https://www.npmjs.com/package/superstatic
> https://www.npmjs.com/package/spa-http-server
> https://github.com/scottcorgan/pushstate-server
FWIW, none of these support push-state _and_ https.

sokolovstas:
> > In the meantime (or if you disagree), any of these SPA focused servers will let you do what you need:
> > https://www.npmjs.com/package/superstatic
> > https://www.npmjs.com/package/spa-http-server
> > https://github.com/scottcorgan/pushstate-server
> 
> FWIW, none of these support push-state _and_ https.

https://www.npmjs.com/package/spa-http-server - works with https and push-state

kimbaudi:
> you can also use `http-server -P http://localhost:8080?` (note the trailing `?`) to proxy your `index.html` file

Please add the note about the trailing `?` to the README. It is important.

PR #236, opened 2016-01-17
Adds ability to proxy requests insecurely

johntron:
This just exposes [`http-proxy`'s `secure` option](https://github.com/nodejitsu/node-http-proxy#options) via `--proxy-secure`. See [README.md](https://github.com/johntron/http-server/blob/ebeeae8bba03f6999e13c00f43d64e0bd8f78d81/README.md#available-options).

Fixes #214

PR #371, opened 2017-05-25
Add 'secure: false' to http proxy

### LINKED ISSUE #214: https certificate error when proxying to server with self-signed certificate
In development we often deal with dev. servers which have self-signed SSL certificate. If I use --proxy option pointing to such a server - I get following error upon proxy attempt:

```
node_modules/http-proxy/lib/http-proxy/index.js:119
    throw err;
          ^
Error: SELF_SIGNED_CERT_IN_CHAIN
    at SecurePair.<anonymous> (tls.js:1381:32)
    at SecurePair.emit (events.js:92:17)
    at SecurePair.maybeInitFinished (tls.js:980:10)
    at CleartextStream.read [as _read] (tls.js:472:13)
    at CleartextStream.Readable.read (_stream_readable.js:341:10)
    at EncryptedStream.write [as _write] (tls.js:369:25)
    at doWrite (_stream_writable.js:226:10)
    at writeOrBuffer (_stream_writable.js:216:5)
    at EncryptedStream.Writable.write (_stream_writable.js:183:11)
    at write (_stream_readable.js:602:24)
```

This will not happen if you pass `secure: false` option to `httpProxy.createProxyServer`. You can do it by default or enable with some option. I can compose a pull request if you like.

tamascsaba:
I use https proxy with self signed cert.

Fixes #214


BigBlueHat:
How do we handle the scenario when not using a self signed cert? This change would effectively mean you could never proxy a secure endpoint securely.

Thoughts?

tamascsaba:
No, ```secure: false``` only means: you don't want to verify the SSL Certs
https://github.com/nodejitsu/node-http-proxy#options

BigBlueHat:
Sure. But there are situations where you *do* want to verify them. Perhaps this should be configurable via an `--proxy-insecure` param?

bdefore:
Thanks @tamascsaba this resolved an issue I had where I was receiving an error of `Error: Hostname/IP doesn't match certificate's altnames` due to it not matching the domain it was supplied for. Certs shouldn't have this issue, but we are transition and need this workaround for development purposes.

I agree with @BigBlueHat that this should have a new flag rather than override by default. Another implementation of this is here: https://github.com/indexzero/http-server/pull/236

thornjad:
I'm going to close this for now in favor of the flag-based approaches in #236 and #459

PR #374, opened 2017-06-09
Remove extra slashes (fixes `-o` command)

jgranick:
The `protocol` variable is set earlier in the code to `"https://"` or `"http://"`, depending on whether SSL is being used. Adding two more slashes results in `"https:////address"` and causes issues on some machines, so this removes the extra slashes :smile:


BigBlueHat:
@jgranick wow...great find. Thanks for the fix!

Issue #381, opened 2017-07-11
why -slient doesn't show the basic info


openks:
is there any way to only show the basic info of the http-server like this:
```js
Starting up http-server, serving ./
Available on:
  http://127.0.0.1:8081
  http://10.10.11.188:8081
Hit CTRL-C to stop the server
```
but no other info.

if use `-s` it shows nothing ,if not it shows all the request info.
if don't us `-s` and request a lot i can't find the basic info of this http-server(if i open more than one http-server) 
if use `-s` i can't finger out which port is working on which http-server.

is anyone can help?thanks

Haroenv:
The point of `-s` is to suppress all output, so there's no way to have only the initial logs, and then hide the other output.

BigBlueHat:
What's the use case for this @openks?

The `-s` options is meant for scenarios where you want absolute silence...so I'm not keen to augment it with an "almost silent" modifier.

You can also specify a specific port to listen on, so I'm not clear on why you'd need to see just the intro output.

Let me know, though. 😄

openks:
yes, you are right ! i can specify a specific port to avoid the question situation @BigBlueHat  thanks

Xmader:
#538

PR #382, opened 2017-07-14
Feature:add parameter `--showserverinfo`

openks:
`--showserverinfo` must use with `-s`, just show serverinfo
more about #381


openks:
just use a  specific port is enough

PR #387, opened 2017-07-28
no need to add it to  array if there is no logFn

clintonyeb:
Will love to receive your feedback and possibly how I can help even if this pull request is rejected. Thanks guys for your hard work, really appreciated!

### DIFF
--- a/lib/http-server.js
+++ b/lib/http-server.js
@@ -58,13 +58,13 @@ function HttpServer(options) {
 
   var before = options.before ? options.before.slice() : [];
 
-  before.push(function (req, res) {
-    if (options.logFn) {
+  if (options.logFn) {
+    before.push(function (req, res) {
       options.logFn(req, res);
-    }
+    })
 
     res.emit('next');
-  });
+  }
 
   if (options.cors) {
     this.headers['Access-Control-Allow-Origin'] = '*';


lib/http-server.js
```diff
@@ -58,13 +58,13 @@ function HttpServer(options) {
 
   var before = options.before ? options.before.slice() : [];
 
-  before.push(function (req, res) {
-    if (options.logFn) {
+  if (options.logFn) {
+    before.push(function (req, res) {
       options.logFn(req, res);
-    }
+    })
```
no1melman: missing semi-colon

lib/http-server.js
```diff
@@ -58,13 +58,13 @@ function HttpServer(options) {
...
 
     res.emit('next');
```
no1melman: This needs to be inside the `before.push(function (req, res) { })` otherwise it will be undefined

BigBlueHat:
Huge thanks for the code review @no1melman!

@clintonyeb would be great to have these tweaks made and the build passing. 😃 Thanks for the work!

no1melman:
@BigBlueHat no problem, just thought I would give my two cents!

thornjad:
Since the source branch has been deleted, changes can't be made on this PR, so there's no path forward. @no1melman please feel free to make a new PR if you'd like!

thornjad:
Reimplemented in #586

Issue #390, opened 2017-08-22
How can I add a `prefix` for public path?


gejiawen:
Hi there,

I want to run a `http-server` instance with a prefix access.

For example,

```shell
$ hs some_folder --prefix /handbook
```

After that, I can access by `http://localhost:8080/handbook` to view the `some_folder`.

How to get this usage?

BigBlueHat:
Do you have another server listening on that port?

If not, then simplest thing would be to rename `some_folder` to `handbook` and run `http-server` in the parent directory.

gejiawen:
@BigBlueHat 

Yeah, I know that. 
I just want to know if `http-server` support a `prefix` configuration. It's seem to do not support.

PR #294, opened 2016-07-19
Add baseDir option, add -b flag to cli options to specify base path

jesseditson:
node-ecstatic allows for a `baseDir` option, which I was looking to use with this lib.
This PR adds the following:
- a `baseDir` option to the main lib file
- a `-b` flag to the cli tool
- a `path` variable to printed/opened URLs
- CLI and README docs for baseDir and flag
- tests for new baseDir option

Fixes #139

Issue #139, opened 2015-03-18
Serve multiple directories


yfr:
Is it possible to serve multiple directories?

marlic7:
+ it will be very useful possibility to configure simple aliases for eg.: /js => ./static/js

Phlogi:
Is there already a way to just serve everything on a subdirectory?

derhuerst:
@Phlogi what do you mean by that?

Phlogi:
like this: domain.com/**myservice**/, i.e. adding a subfolder.

derhuerst:
AFAIK, it's not possible with *http-server* out of the box.*ecstatic*, the core behind *http-server*, provides an [option `baseDir`](https://www.npmjs.com/package/ecstatic#opts-basedir).

Issue #63, opened 2014-01-09
allow different cache times for different urls


lanterndev:
Currently http-server allows you to configure the "cache-control: max-age" value for all urls. But often you want a short max-age for e.g. /{index.html} and a long max-age for urls to versioned assets like /styles/f35ac111.main.css.

Would you accept a patch to allow specifying different max-ages for different urls? (This could be implemented most generally by matching the request url against one or more configurable regexes which map to corresponding max-ages, for instance. But if it's simpler, just giving the long cache time to everything that isn't in the root directory, or every asset with a name like a1b2c3d4.filename.ext would probably work for many use cases.)

kilianc:
+1 on this. Multiple `-c` args should be supported. I would be ok with just not setting cache for `.html` files but since the implementation is very simple we could probably aim to have something like:

    $ http-server -c31536000:**/*.css -c-1:**/*.html

?

kilianc:
I would be ok PRing to this if a PR is welcome.

derhuerst:
I'm not sure this is possible in the scope of `http-server`.

First of all, "http-server is a *simple* […] command-line http server". I think it's sad that JS projects on GitHub trying to keep things simple (and leave advanced use cases to others) are constantly made more complex. But that's just MHO.

Second, `http-server` is just a layer on top of [ecstatic](https://github.com/jfhbrook/node-ecstatic), which seems to allow *only one* caching duration.

kilianc:
> Second, http-server is just a layer on top of ecstatic, which seems to allow only one caching duration.

@derhuerst the issue on ecstatic is right here in this thread, with a merged PR to add support to custom caching.

> First of all, "http-server is a simple […] command-line http server". I think it's sad that JS projects on GitHub trying to keep things simple (and leave advanced use cases to others) are constantly made more complex. But that's just MHO.

It still seems reasonable to me that a static assets server has advanced support for caching. Adding this has not impact on who's not gonna have a use for multiple caching values.

Also, what do you think is going to happen if this doesn't land? People are going to use `ecstatic` or `connect` and roll out their own server.

derhuerst:
@kilianc 

>> Second, http-server is just a layer on top of ecstatic, which seems to allow only one caching duration.
>
> @derhuerst the issue on ecstatic is right here in this thread, with a merged PR to add support to custom caching.

Sorry, didn't see this.

[...]

PR #395, opened 2017-09-05
Add baseDir option and pass-through to ecstatic

idmontie:
Upgrade ecstatic to version 3.0.0. Use ecstatic's baseDir option
to set a base directory.

See #390

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,8 @@ This will install `http-server` globally so that it may be run from the command 
 
 `-a` Address to use (defaults to 0.0.0.0)
 
+`-b` Base path to serve files from (defaults /)
+
 `-d` Show directory listings (defaults to `true`)
 
 `-i` Display autoIndex (defaults to `true`)
--- a/bin/http-server
+++ b/bin/http-server
@@ -18,6 +18,7 @@ if (argv.h || argv.help) {
     'usage: http-server [path] [options]',
     '',
     'options:',
+    '  -b           Base path [/]',
     '  -p           Port to use [8080]',
     '  -a           Address to use [0.0.0.0]',
     '  -d           Show directory listings [true]',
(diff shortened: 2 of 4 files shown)


BigBlueHat:
Thanks for this PR, @idmontie.

One concern, though, is that `http-server` will still tie up whatever pot it's assigned to (i.e. 8080), but won't be able to serve anything else except what's under `baseDir`.

I fear it would encourage the wrong exceptions, such as "oh good! now `http-server` can serve my `/blog/` directory, and my other apps can serve everything else!"--which would not be the case...

What are your thoughts on avoiding that misconception?

BigBlueHat:
#139 and #63 also feel related and worth a read.

idmontie:
@BigBlueHat That's true, I didn't think about that.

In my case, I'm emulating a production folder structure on my local system, so files are usually pushed up to a subdirectory on a server, while on my local machine, they simply live in a build folder.

I can see where the confusion lies. Maybe I can update the documentation/options docs to reflect that the port will be used, even though the files may be served with a prefix.

BigBlueHat:
That use case makes sense. If you can make that clear enough in the docs, then this is probably worth having--certainly others face the same scenario.

Thanks!

hikumealan:
Any ETA on this? I have the same use case and would really like to see this feature.

yashha:
Thats good if you want to test the page routing when you know it will be served on a subpath in the futuer.

kolisko:
+1

thornjad:
Alright, @idmontie's #395 and @jesseditson's #294 have a one-line difference. I'm going to go with #294 because it came first

MaximSagan:
> Alright, @idmontie's #395 and @jesseditson's #294 have a one-line difference. I'm going to go with #294 because it came first

@thornjad This issue was closed, but neither of the PRs you mentioned were ever merged. As #294 (the one you said you'd go with) has gone stale with merge conflicts since then, I have rebased the it into #837.

hairinwind:
> 

Can this be merged? We do need this feature. Thanks!

PR #403, opened 2017-10-18
Add npm package-lock

thornjad:
This adds `package-lock.json` which will help prevent future dependency problems. From the [npm documentation](https://docs.npmjs.com/files/package-locks):

> This file describes an exact, and more importantly reproducible `node_modules` tree. Once it's present, any future installation will base its work off this file, instead of recalculating dependency versions off `package.json`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "http-server",
-  "version": "0.10.0",
+  "version": "0.11.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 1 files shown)


indexzero:
Should we be publishing `package-lock.json` files? If `http-server` is used as a dependency instead of a global CLI tool wouldn't this complicate the install tree?

Raindeer44:
If published, it would definitely end up complicating the install tree of anything using it as a dependency. However, part of the beauty of `package-lock.json` is that it _can't_ be published. When publishing, npm will simply ignore it. It's designed for use in the source code only, i.e. here on GitHub.

Raindeer44:
Using a `package-lock.json` ensures that everyone who works with this source code will be using the exact same versions of every dependency. Its use is strongly encouraged by npm.

BigBlueHat:
@thornjad we should try and get `package-lock.json` updated and committed before the next release. It is, as you've noted, the new normal. 😃

thornjad:
Perfect! I just updated this PR so the `package-lock.json` is up to date with `master`.

BigBlueHat:
Thanks @thornjad!

PR #404, opened 2017-10-23
Enable customizable cache time

### LINKED ISSUE #63: allow different cache times for different urls

houston3:
Ecstatic supports passing a function as a cache argument, but http-server passes the argument as a string. Added eval to make the function work. Also updated tests.

Fixes #63

### DIFF
--- a/bin/http-server
+++ b/bin/http-server
@@ -35,7 +35,7 @@ if (argv.h || argv.help) {
     '  -o [path]    Open browser window after starting the server.',
     '               Optionally provide a URL path to open the browser window to.',
     '  -c           Cache time (max-age) in seconds [3600], e.g. -c10 for 10 seconds.',
-    '               To disable caching, use -c-1.',
+    '               To disable caching, use -c-1. Also supports a cache function, which can be an inline function or a path to an external module. https://github.com/jfhbrook/node-ecstatic#optscache',
     '  -t           Connections timeout in seconds [120], e.g. -t60 for 1 minute.',
     '               To disable timeout, use -t0',
     '  -U --utc     Use UTC time format in log messages.',
--- a/test/http-server-test.js
+++ b/test/http-server-test.js
@@ -379,5 +379,105 @@ vows.describe('http-server').addBatch({
     teardown: function (server) {
       server.close();
     }
-  }
+  },
+  'When cache time is enabled': {
+        topic: function () {
+          var server = httpServer.createServer({
+            root: root,
+            cache: '10'
+          });
+          server.listen(8086);
+          this.callback(null, server);
+        },
+        'and given GET request': {
+          topic: function () {
+            request('http://127.0.0.1:8086/file', this.callback);
+          },
+          'cache max-age should be 10': function (res) {
+            assert.equal(res.headers['cache-control'], 'max-age=10');
+          }
+        },
+        teardown: function (server) {
+          server.close();
+        }
+      },
+  'When inline cache function is enabled': {
+      topic: function () {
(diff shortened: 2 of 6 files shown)


bin/http-server
```diff
@@ -92,10 +92,11 @@ else {
   listen(port);
 }
 
+/* jshint evil:true */
 function listen(port) {
   var options = {
     root: argv._[0],
-    cache: argv.c,
+    cache: eval(argv.c),
```
thornjad: While this definitely works to pass a function through (which is good!), using `eval` runs into an issue when `argv.c` is meant to be a string like `'max-age=120'`. While that string is a perfectly valid cache header, it is not a valid JS expression, so `eval` will error.

test/http-server-test.js
```diff
@@ -154,5 +154,49 @@ vows.describe('http-server').addBatch({
         assert.ok(res.headers['access-control-allow-headers'].split(/\s*,\s*/g).indexOf('X-Test') >= 0, 204);
       }
     }
-  }
+  },
+  'When cache time is enabled': {
```
thornjad: We need cache time as a string to round out this suite

houston3:
Added support for string args.

peschee:
Are there any updates on this? Is something missing in order to merge this?

bin/http-server
```diff
@@ -92,10 +92,16 @@ else {
   listen(port);
 }
 
+/* jshint evil:true */
 function listen(port) {
+  var cache = argv.c;
+  try {
+    cache = eval(argv.c);
+  }
+  catch (e) {}
```
thornjad: Should we report to the user that their function failed?

bin/http-server
```diff
@@ -29,7 +29,7 @@ if (argv.h || argv.help) {
     '                     Optionally provide CORS headers list separated by commas',
     '  -o [path]    Open browser window after starting the server',
     '  -c           Cache time (max-age) in seconds [3600], e.g. -c10 for 10 seconds.',
-    '               To disable caching, use -c-1.',
+    '               To disable caching, use -c-1. Also supports a cache function. https://github.com/jfhbrook/node-ecstatic#optscache',
```
thornjad: We should get this added to the README too

bin/http-server
thornjad: If we get into the [empty] catch block, what will cache be set to? Are we certain it will still be equal to `argv.c`? Should we use an intermediate variable?

thornjad:
You'll need to merge or rebase master, your branch doesnt run tests on Node 12 or 13 due to an outdated `package.json` loading in an old version of `vows`

bin/http-server
houston3: Added logging.

bin/http-server
houston3: Updated README.

bin/http-server
houston3: Added intermediate variable.

thornjad:
Closing as out-of-scope, see https://github.com/http-party/http-server/issues/63#issuecomment-879124467

PR #409, opened 2017-11-26
make http-server extensible through plugins

amoldavsky:
this change brings:
- events to http-server
- extensibility via plugins
- externalization of existing functionality into plugins
  - http-server-cors
  - http-server-logger
  - http-server-robots
- a missing test for `options.logFn`


BigBlueHat:
@amoldavsky thanks for working on this! I'd love to see some docs about how to write new plugins and a general explanation of how this approach works.

I'd also want to be very sure that the "out of the box" functionality stays intact--as that's our primary use case and target audience: "a simple zero-configuration command-line http server"

I do like the idea of a more modular codebase and more configurable deployment options. Thanks for hacking on this idea!

amoldavsky:
@BigBlueHat hey thanks for the comments!
Not a problem, I can provide documentation on the plugins, will create a README in the `/plugins` dir.

"out of the box" functionality remains as is, I believe the unit tests covered most of it ( I added one that was missing ) and all passing. We can add more tests as we go of course.

avioli:
http-saber?

amoldavsky:
@avioli  don't like the name? :)))
https://github.com/amoldavsky/http-saber

PR #410, opened 2017-12-02
Integrate TravisBuddy

bluzi:
TravisBuddy will comment on pull requests in your public repository everytime a test failed in one of them. The comment will include only the part of the build log that applies to your testing framework, so your contirbutors won't have to enter Travis's website and search the long and annoying build log for the reason the tests failed.

Here's an example: bluzi/static-server#1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,3 +11,13 @@ before_install:
 
 script:
   - npm test
+
+notifications:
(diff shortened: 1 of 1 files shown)


BigBlueHat:
I think we'll stick with GitHub's Travis integration, but thanks for the interest!

PR #415, opened 2017-12-19
Add optional extra middleware support

whatl3y:
I have a small static website that I'm using http-server to serve, but wanted to do some extra logging to a database for each request that was made.

I didn't see an option to add additional code before each request is responded to, so added a new parameter to supply a file path that exports a middleware function that can be executed with each request.


thornjad:
While I like the concept of being able to plug arbitrary additional middleware into http-server (thus making it far more extensible), I'm not sure I like the idea of adding it to the CLI. I think this should be part of a wider push to make a better API for http-server which can be `require`'d and `import`'ed. This is something I'm looking at as I explore rewriting http-server.

PR #423, opened 2018-01-19
Proxy error handling

sjohnsonaz:
I added a small bit of error handling to the proxy section.  It causes the server to return a 503 status, rather than exit.

### DIFF
--- a/lib/http-server.js
+++ b/lib/http-server.js
@@ -1,10 +1,10 @@
 'use strict';
 
 var fs = require('fs'),
-    union = require('union'),
-    ecstatic = require('ecstatic'),
-    httpProxy = require('http-proxy'),
-    corser = require('corser');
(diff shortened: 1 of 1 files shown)


apixandru:
I'm having this exact same problem now, the server just crashes with regular chrome usage.
If the http server proxies /request1 to http://otherserver/request and i go and access /request1 and refresh a couple of times without letting the request go through, the http-server crashes.

rafaelsales:
It feels like the owner has to add more maintainers to the project so that it scales, or it will become a ghost town

BigBlueHat:
This generally looks good. It needs some tests and the whitespace changes removed. Once those are added, this seems like a good addition. Thanks!

thornjad:
Closing in favor of the non-crashing logging in #477

PR #434, opened 2018-03-22
Add support for credentials requests

senaev:
To serve files with credentials support cross domains we ned to not just add some headers in request, we should also change Access-Control-Allow-Origin header to Origin value from request.

I added a new flag for it, because is is different from --cors logic and may be not safe for users.

### DIFF
--- a/test/http-server-test.js
+++ b/test/http-server-test.js
@@ -152,6 +152,43 @@ vows.describe('http-server').addBatch({
       },
       'response Access-Control-Allow-Headers should contain X-Test': function (err, res) {
         assert.ok(res.headers['access-control-allow-headers'].split(/\s*,\s*/g).indexOf('X-Test') >= 0, 204);
+      },
+      'response Access-Control-Allow-Credentials should not be presented': function (err, res) {
+        assert.equal(res.headers['access-control-allow-credentials'], undefined);
+      }
+    }
+  },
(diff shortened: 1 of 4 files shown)


senaev:
@indexzero this is for dev purposes

thornjad:
@senaev would you mind merging/rebasing master?

senaev:
> would you mind merging/rebasing master?

@indexzero @thornjad done

test/http-server-test.js
```diff
@@ -155,6 +155,43 @@ vows.describe('http-server').addBatch({
...
+    }
+  },
+  'When credentials is enabled': {
+    topic: function () {
+      var server = httpServer.createServer({
+        root: root,
+        credentials: true,
+        corsHeaders: 'X-Test'
+      });
+      server.listen(8083);
```
thornjad: This port is already being used by username/password tests, so we should do 8085 here (8084 is also claimed). This is causing the tests to fail.

senaev: 👌

senaev:
@thornjad check this please

velzie:
I'm confused. From my understanding, `Access-Control-Allow-Credentials: true` is set on the target server of an api request so that it can receive cookies, but this is a static http server which never reads sent cookies. Can you elaborate more on what your use case for this is?

Issue #162, opened 2015-06-22
Support PEM passphrase for HTTPS


tehsenaus:
See `passphrase` option here:
https://nodejs.org/api/tls.html#tls_tls_createserver_options_secureconnectionlistener

chris--jones:
Can I take this one?

PR #437, opened 2018-03-24
Feat passphrase support

kiranghule:
Added support for providing passphrase to decrypt private key

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,8 @@ This will install `http-server` globally so that it may be run from the command 
 
 `-K` or `--key` Path to ssl key file (default: key.pem).
 
+ `-P` or `--passphrase` Pass phrase to decrypt the key (optional).
+
 `-r` or `--robots` Provide a /robots.txt (whose content defaults to 'User-agent: *\nDisallow: /')
 
 `-h` or `--help` Print this list and exit.
--- a/bin/http-server
+++ b/bin/http-server
@@ -37,6 +37,7 @@ if (argv.h || argv.help) {
     '  -S --ssl     Enable https.',
     '  -C --cert    Path to ssl cert file (default: cert.pem).',
     '  -K --key     Path to ssl key file (default: key.pem).',
+    '  -P --passphrase     Pass phrase to decrypt the key (optional)',
     '',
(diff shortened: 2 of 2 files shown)


kiranghule:
Closes [https://github.com/indexzero/http-server/issues/162](url)

BigBlueHat:
What's the use case for this? It seems unwise to provide a passphrase in clear text on the command line. Happy to hear more, though.

BigBlueHat:
Thanks @kiranghule. Guess I'm just hesitant to encourage passphrases exposed on the command line. Perhaps an environment variable (if present) could be used for this scenario? Certainly Heroku (and friends) would store those in a safer place than in a clear text command line request.

💭's?

kiranghule:
@BigBlueHat Agreed. It's better to get this from an environment variable. Will look into it.

thornjad:
@kiranghule are you still interested in moving forward with this PR?

kiranghule:
> @kiranghule are you still interested in moving forward with this PR?

Will look into this, this weekend. The last conclusion I guess was to pass the passphrase via the env variable. But I have some doubts regarding how to go about it. Any help would be appreciated. Thanks.

thisIsTheFoxe:
Hey, great idea.. 

but does that even work? Because for me it didn't.. just as @thornjad mentioned in #456, bc [union](https://github.com/flatiron/union/blob/6ece680c3c4e2480b74250701b54da6bb61b44d2/lib/core.js#L72..L75) ignores the passphrase anyway..
<hr>

In our use case we would start the server manually each time, so I think passing the passphrase via CLI should at least be possible.
However, the use of env vars would be pretty straightforward, wouldn't it? Similar to [NODE_HTTP_SERVER_PASSWORD](https://github.com/http-party/http-server/blob/d7bce39827d4db03190cb33b5f33d3a543464ee1/bin/http-server#L131)

README.md
```diff
@@ -59,6 +59,8 @@ This will install `http-server` globally so that it may be run from the command
 
 `-K` or `--key` Path to ssl key file (default: key.pem).
 
+ `-P` or `--passphrase` Pass phrase to decrypt the key (optional).
```
cor3000, line 62: seems to conflict with the option **-P / --proxy**

Issue #280, opened 2016-05-25
Is it possible to proxy some url to other server not all url.


kanbekotori:
/a.html -> http-server
/api/data -> other-server

alifatic:
anybody know this answer?

Keksike:
Hey @kanbekotori & @alifatic 

I solved this issue with a new `-U` option: https://github.com/indexzero/http-server/pull/439

thornjad:
I'm inclined to say this is out of scope for our simple http-server, would a use case like this be better suited to a fully-featured server like Apache?

Issue #377, opened 2017-06-21
Implement Firebase-style rewrite/redirect/header JSON config


BigBlueHat:
So far we've avoided doing anything out of config files. However, there's a growing interest in being able to add additional headers, redirect things, and rewrite certain URL's before they hit the file system.

PR #365 has some code for doing such a thing with a (currently) unique/custom JSON format. The discussion there led to wanting (and then finding) a couple rewrite-focused JSON formats. The most capable (so far) is the one from Firebase. See some of that discussion at
https://github.com/indexzero/http-server/pull/365#issuecomment-307281707

So. Some questions:
 - does using a config file "muddies" this project's purity as a command line "only" server?
 - are there other (more) standard rewriter formats we should consider?
 - are there non-config file based options for rewriting and redirecting similar to the `.headers` propose in #360?

Having this system *would* expand the use cases for this server, so it's a bit of a pivot point project-wise. It also has performance impact that will need consideration/optimization.

Thoughts? 😄

PR #439, opened 2018-04-05
Proxy path option

cibelius:
Add new "proxy path" `-R` option, which you can use to specify a certain path for which the requests are proxied. Other paths are not proxied.

Example:

If I want to proxy only requests done to path `/api/`, I can set `-R /api/`. Then:

GET /a.html -> not proxied
GET /api/data -> proxied

Solves issue: https://github.com/indexzero/http-server/issues/280

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ This will install `http-server` globally so that it may be run from the command 
 
 `-P` or `--proxy` Proxies all requests which can't be resolved locally to the given url. e.g.: -P http://someurl.com
 
-`-U` Path for which proxy is enabled. e.g.: -U /api/ (defaults to '')
+`-R` Path for which proxy is enabled. e.g.: -U /api/
 
 `-S` or `--ssl` Enable https.
 
--- a/README.md
+++ b/README.md
@@ -53,6 +53,8 @@ This will install `http-server` globally so that it may be run from the command 
 
 `-P` or `--proxy` Proxies all requests which can't be resolved locally to the given url. e.g.: -P http://someurl.com
 
+`-U` Path for which proxy is enabled. e.g.: -U /api/ (defaults to '')
+
 `-S` or `--ssl` Enable https.
 
 `-C` or `--cert` Path to ssl cert file (default: cert.pem).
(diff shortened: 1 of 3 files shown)


Keksike:
Hey,

I guess this will still need some work. I works perfectly fine when testing locally, but on my AWS deployment it somehow breaks.

Keksike:
Actually I guess it works, issue is probably just with my AWS setup :)

BigBlueHat:
Thanks for coding this idea!

First, some minor house keeping:
 - squash the current commits with focused commit messages
 - write tests!

Second, we'd need to address some of the confusion around the semantics of the command and the *likely* request for doing this with multiple paths. URL rewriting has come up before and #377 explores a "config-based" approach. However, keeping `http-server` "config-less" (or command line switches only) is also *very* appealing.

Consequently, things could get confusing when folks try to figure out how to use this feature. For instance:
 - `http-server --proxy http://api.example/` - works now; everything that isn't on disk goes here
 - `http-server --proxy http://api.example/ -R /api/` - this PR; only things not on disk *and* that contain this prefix
 - `http-server -R /api/` - would fail; should it throw an error? just get ignored?
 - `http-server --proxy http://api.example/ -R /api/ -R /a/` - not sure what would/should happen here...
 - `http-server --proxy http://api.example/ -R /api/ --proxy http://db.example/-R /db/` - ...someone will ask for this...

Currently, the `--proxy` option will catch *everything* not on disk...and I can understand not wanting that to happen. However, because of the scenarios mentioned above, providing this `-R` option (which might need renaming...) would sadly add confusion, not clarity. But, I think this has merit! It just needs some more discussion/code to make it sing. 😃 

There are bound to be other servers out there who handle a "path + URL" mapping from the command line. Perhaps picking up some of their tips and trips may help here.

Keksike:
https://github.com/txchen/light-server has a similar kind of implementation as the one I did (`--proxypath <proxypath`). I found it perfect for my needs.

It also has a optional config file, which might be a good way to implement the configuration (e.g. for various proxypaths) which was mentioned above.

one-github:
This was almost working, except ``res.emit('next');`` was missing:

i3ene:
I personally find the idea of a more complex configuration file good. There is also a good approach with #906 with the specified example.
Doing everything over cli parameters gets confusing pretty fast.

PR #447, opened 2018-05-31
documented the url option of -o.

intotecho:
The description in README.md was missing this option that apears in the -h output.
Also, by adding http:// to the example, it hints to the user that this is required in the windows shell, otherwise the url is treated as another command


thornjad:
Moved to #514 since this fork/branch no longer exist.

PR #448, opened 2018-06-05
Upgrade ecstatic from 3.0.0 to 3.2.0

tunylund:
Ecstatic 3.2.0 uses a newer version of 'mime' and thus provides support
for javascript modules (.mjs) files to be served with the approppriate
content-type.

ES6 import statements work in the browser only if the server responds
with 'application/javascript' content type.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
   "dependencies": {
     "colors": "1.0.3",
     "corser": "~2.0.0",
-    "ecstatic": "^3.0.0",
+    "ecstatic": "^3.2.0",
     "http-proxy": "^1.8.1",
     "opener": "~1.4.0",
     "optimist": "0.6.x",


BigBlueHat:
I plan to bump the `node-ecstatic` version once they've moved up `mime` to `^2.0.0`. Ideally this will happen along with #276 and exposing `node-ecstatic`'s `--mime-types` parameter also.

khoin:
This seems to have been resolved? I was working today and it complained about the wrong MIME type so I did `npm install http-server -g` again which, afterwards, works perfectly.

BigBlueHat:
@khoin it's possible the `.mjs` file extension's been mapped via some minor point release in recent `mime` versions. I'll close this--as we'll be doing ecstatic upgrades along with other `http-server` point releases. If this specific issue rises again, feel free to file an issue for it and we'll look at the various options to fix it.

Cheers!

PR #452, opened 2018-07-14
Prevent vows from swallowing test errors on uncaught errors within topics

IamfromSpace:
Hello, been looking around at making some further contributions, and came across an issue where vows would swallow errors, making it very difficult to debug.  Very simple one liner (with a comment to explain its purpose).

What was a fairly simple issue (I had a process listening on 8081 that I'd forgotten about), prevented my tests from passing, with no meaningfully related error output.  This cost me a little over an hour of troubleshooting, until I was able to find the magic `uncaughtException` event, which clearly spelled out the problem.  Hoping to contribute back to help the next person!

Validating the behavior difference is fairly simple, just start an http-server on 8081 before executing tests on master vs this branch.

Cheers!

### DIFF
--- a/test/http-server-test.js
+++ b/test/http-server-test.js
@@ -5,6 +5,9 @@ var assert = require('assert'),
     request = require('request'),
     httpServer = require('../lib/http-server');
 
+// Prevent vows from swallowing errors
+process.on('uncaughtException', console.error);
+
 var root = path.join(__dirname, 'fixtures', 'root');
 
 vows.describe('http-server').addBatch({


BigBlueHat:
Thanks for this change!

@IamfromSpace is this mostly to get ride of the `✗ Errored » callback not fired` messages (which are rather confusing)?

It might be good if we also threw an error/warning about something already listening on `8080` and/or `8081`--as something on either of those ports will make some mix of the tests fail.

Thoughts?

IamfromSpace:
Yep, that's exactly the error I was seeing!

As for more explicit error messaging around a port in use (as absolutely, that should be one of the most common cases), once the error got through, that's pretty much what it said.  After having been totally stumped for a good long while it was immediately clear to me what was happening.  I'd opt for a simpler implementation, given that the error is pretty clear.

PR #453, opened 2018-07-17
Fix preferGlobal property in package.json

YangFan789:
preferGlobal should be an boolean


BigBlueHat:
Looks like `preferGlobal` has been deprecated, so not sure it matters much--and at some point we should remove it:
https://docs.npmjs.com/files/package.json#preferglobal

YangFan789:
okay

PR #460, opened 2018-08-10
Dynamic response headers

jimlawhorn:
#360 

Implements a similar concept to the two referenced implementations.  Including an optional headers.json allows for adding any response header based on a requested path and multiple matches can be appended together.  It can also be used to override command-line headers for specific paths.

Major differences 
- Headers are stored in a json file at the root and is read on startup, not on request.  This limits to a single read.
- Implemented globbing (minimatch), which should allow for more robust pattern matching e.g being able to specify `*.js` rather than a config at a directory or specific file level.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,10 +63,42 @@ This will install `http-server` globally so that it may be run from the command 
 
 `-h` or `--help` Print this list and exit.
 
+## Headers
+
+An optional file called `headers.json` may be placed in the root path (as specified by the [path] argument) being
+served.  If this file is present, the contents are used to add headers to the server response.  Request paths can be
+added to the json file with an array of objects that contain information about the headers to add.
+
+The `headers.json` file follows this format:
+```
+{
+  "Path to Match": [
+    {"header": "Name of the header to add", "value": "value of header", "append": true/false},
+    {"header": "Second header on same path", "value": "value of header", "append": true/false}
+  ],
+  "Second Path to match": [
+    {"header": "Name of the header to add", "value": "value of header", "append": true/false}
+  ]
+}
+```
+
+Notes about the `headers.json` file:
+- "Path to match" uses [minimatch](https://www.npmjs.com/package/minimatch) to match to the request.  For example,
+  an entry for `/**` would match all requests.
+- The headers are added in the order they appear in the file.  Unless `append` is set to true, previous values will
+  be overwritten
+- `append` is optional.  If not present, or set to false, a matched header will be overwritten.  If set to true,
+  the value specified will be appended to the existing header value separated by a semi-colon (;) and a space as
+  specified [here](https://tools.ietf.org/html/rfc7230#section-3.2).
+- Headers added with this file will be added/set AFTER headers set by command line arguments.  This will allow you
+  to selectively override/change headers set at http-server startup.
+
+
 ## Magic Files
 
 - `index.html` will be served as the default file to any directory requests.
 - `404.html` will be served if a file is not found. This can be used for Single-Page App (SPA) hosting to serve the entry page.
+- `headers.json` will be read (if present) and used to add headers to the response.
 
 # Development
 
--- a/lib/http-server.js
+++ b/lib/http-server.js
@@ -6,6 +6,7 @@ var fs = require('fs'),
     httpProxy = require('http-proxy'),
     corser = require('corser');
 
+var HEADERS_FILE_NAME = 'headers.json';
 //
 // Remark: backwards compatibility for previous
 // case convention of HTTP
@@ -89,6 +90,27 @@ function HttpServer(options) {
         return res.end(robots);
       }
 
+      res.emit('next');
+    });
+  }
+
+  if (fs.existsSync(this.root + '/' + HEADERS_FILE_NAME)) {
+    console.log('Loading Headers file:', this.root + '/' + HEADERS_FILE_NAME);
+    var headers = JSON.parse(fs.readFileSync(this.root + '/' + HEADERS_FILE_NAME)) || {};
+    var minimatch = require('minimatch');
+    var paths = Object.keys(headers);
+    before.push(function (req, res) {
+      paths.forEach(function (p) {
+        if (minimatch(req.url, p, {})) {
+          headers[p].forEach(function (h) {
+            var value = h.value.trim();
+            if (h.hasOwnProperty('append') && h.append) {
+              value = res.getHeader(h.header).trim() + '; ' + value; // Trim .getHeader() as we may not have set it
+            }
+            res.setHeader(h.header, value);
+          });
+        }
+      });
       res.emit('next');
     });
   }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,2399 @@
+{
+  "name": "http-server",
+  "version": "0.11.1",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "JSV": {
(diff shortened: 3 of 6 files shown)


jimlawhorn:
Would someone mind kicking off the build again?  I can't figure out the issue.  It currently runs fine for node version 4 (failed build version) on my local (MacOS), so I'm wondering if it's a build problem.

Local build log:
```
http-server (dynamic_headers) $ npm --version
2.15.11 http-server (dynamic_headers) $ clear

http-server (dynamic_headers) $ node --version
v4.9.1
http-server (dynamic_headers) $ npm --version
2.15.11
http-server (dynamic_headers) $ nvm --version
0.33.11
http-server (dynamic_headers) $ npm install
npm WARN prefer global jshint@2.9.5 should be installed with -g
npm WARN prefer global jsonlint@1.6.2 should be installed with -g
http-server (dynamic_headers) $ npm test

> http-server@0.11.1 pretest http-server
> common bin/http-server lib/ test

Running jshint with options: -c common-style/dist/.jshintrc bin/http-server lib/ test
Running jscs with options: -c common-style/dist/.jscsrc bin/http-server lib/ test

> http-server@0.11.1 test http-server
> vows --spec --isolate

 
  ♢ http-server 
  
  When http-server is listening on 8080 when headers.json is present
    ✓ should be readable
  When cors is enabled and given OPTIONS request
    ✓ status code should be 204
    ✓ response Access-Control-Allow-Headers should contain X-Test
  When http-server is listening on 8080 it should serve files from root directory
    ✓ status code should be 200
  When http-server is listening on 8080 when requesting non-existent file
    ✓ status code should be 404
  When http-server is listening on 8080 When http-server is proxying from 8081 to 8080 it should serve files from the proxy server root directory
    ✓ status code should be the endpoint code 200
  When http-server is listening on 8080 when requesting /
    ✓ should respond with index
  When http-server is listening on 8080 when robots options is activated
    ✓ should respond with status code 200 to /robots.txt
  When http-server is listening on 8080 when headers.json is present when requesting /
    ✓ should respond with custom header
  When http-server is listening on 8080 and options include custom set http-headers
    ✓ should respond with headers set in options
  When http-server is listening on 8080 it should serve files from root directory and file content
    ✓ should match content of served file
  When http-server is listening on 8080 When http-server is proxying from 8081 to 8080 it should serve files from the proxy server root directory and file content
    ✓ should match content of the served file
  When http-server is listening on 8080 When http-server is proxying from 8081 to 8080 it should fallback to the proxied server
    ✓ status code should be the endpoint code 200
  When http-server is listening on 8080 When http-server is proxying from 8081 to 8080 it should fallback to the proxied server and file content
    ✓ should match content of the proxied served file
(node) util.print is deprecated. Use console.log instead.

✓ OK » 14 honored (0.088s) 
```

BigBlueHat:
I do think the top-level file is probably going to handle the most cases. However, I do like the "in place" `.header` approach as it avoids the need to update a "parent" file and/or restart the server when changes happen within the site. Thoughts?

jimlawhorn:
I think there are pros and cons to both approaches (in directory vs global).  

From my perspective, probably the hardest problem to solve from an in directory model is the inheritance aspect of it.  If a requested file is several directories deep and each directory has a .headers file, how do you determine which to apply?  While most standard headers only allow for a single value, that's not necessarily true of custom headers.  If we were to say that lowest-level (e.g. closest to the file requested) "wins" and headers aren't appended as you go up the chain, it turns into a potential maintenance issue where you're setting the same headers in multiple .headers files.  If we were to go the other way and allow files up the chain to append, researching where a header came from would be difficult - and that assumes we knew which could be appended vs which must be a single value.

If I'm understanding at least one of your concerns correctly, it's the need for a restart on changes.  I think if I added a file watcher to the global header file so that it got reloaded when updated, it would address that concern.  If you were to add new files to the site, you would either be okay with the existing rules and no header changes were necessary, or you would add headers to the global - at which point a reload would occur.

Having said all that, I really think we should support both (in-place and global), but I would propose the following:
- In-place only applies to the directory it's in, not child directories.  There is no inheritance.
- Once a header is set, it is not overridden by the global.
- The global will append headers if the append setting is on.
- Directory .headers are checked on each request.
- Global header is updated on startup or based on a file watcher.

Taking the combined approach:
- Allows you to be very specific when necessary and not pollute the global with a bunch of one-offs.  
- It still allows for the global to prevent additional maintenance (headers file in every directory for `.js`).
- Simplifies trouble-shooting.  The header is either coming from the directory or from the global.

What do you think?

BigBlueHat:
@jimlawhorn thanks for reasoning through all that! Inheritance wasn't on my personal list of "wants" for this feature. The whole idea came from the W3C's Web Platform Tests server--used to test all the browsers. They have several different "handlers" one of which is the File Hanlder:
https://wptserve.readthedocs.io/en/latest/handlers.html#file-handlers

That describes essentially what I'd personally like to see. Essentially, the "most" it does is check three total files *at request time* (so no "global" and no inheritance):
```
/example-folder/
 - __dir__.headers
 - index.html.headers
 - index.html
```

A request for `/example-folder/` or `/example-folder/index.html` or `/example-folder/index` would result in all three of those files being read. However, they're minimal reads with no parsing as the `.headers` files are "raw" HTTP headers. So, it's essentially a raw concatenation of `__dir__.headers` and `index.html.headers` + the response body being `index.html` (also unparsed/processed). Additionally, the `__dir__.headers` file only effects it's sibling files--i.e. it doesn't "cascade" into sub-directories (so no inheritance in that sense).

What this allows is for multiple directories from multiple sources to be tossed into one web space and served without a "global" file being consulted and/or changed to deal with the new scenarios. Each directory is "self sovereign" in that sense.

It also avoids additional file monitoring, reloading, etc as everything is read at request time.

It's sort of a [Principle of Least Astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) thing. 😃 It means you can analyze each requests potential response output simply by reading these 1-3 raw sibling files--vs. consulting, parsing (mentally or otherwise) a file elsewhere in the hierarchy.

It is certainly a more limited approach, but it also keeps `http-server` dumb by design. 😄 Something I find valuable for longevity and stability.

Thoughts?

jimlawhorn:
I would argue that Principle of Least Astonishment would indicate that we should follow what other popular web servers are doing.  Taking cues from Apache HTTP server, Nginx, Netlify, etc. would lead to a single configuration file at the root (or virtual host) that is only read at server start.

While some implementations do allow for directory specific configurations (like Apache's `.htaccess`), it is generally [not advised](https://httpd.apache.org/docs/trunk/howto/htaccess.html).  Nginx goes so far as to say [it's wrong](https://www.nginx.com/resources/wiki/start/topics/examples/likeapache-htaccess/)

I think there are use cases for both, which is what I'm advocating.  I only have use cases for the dynamic headers at the root level.  I'm happy to _also_ implement the directory/file-level headers that you have described, but implementing _only_ directory/file-level headers is not something that I'm interested in contributing. 

Thoughts?

thornjad:
Sorry to jump in after so much time, I agree that a directory-level configuration is best long-term. However, least astonishment for me makes me want to call the file `.headers`. The dot ensures it's _usually_ hidden, because it's not really a "real" file to be served.

I also like reading on startup, rather than on request.

I'm unsure about the JSON format and whether it introduces more complexity than a "simple http server" should offer? I'm not entirely convinced in any direction.

In the case we do want the added complexity, what would you think about a more powerful and less verbose format like YAML or something else?

Issue #445, opened 2018-05-16
brotli support


mendrik:
Would be a nice to have, like --gzip

[Brotli](https://github.com/google/brotli)

BigBlueHat:
Definitely open to PRs for this one. 😄

thornjad:
This is definitely needed! I've just started working on adding brotli support in the upstream project ecstatic first, and I'll open a PR here once ecstatic releases it. My thought is to be able to pass it just like gzip, i.e. `http-server --brotli`

thornjad:
See jfhbrook/node-ecstatic#228

PR #152, opened 2015-05-13
support custom mime-type(s) via .types file

dotnetCarpenter:
This is my take on fixing #35, if [my upstream PR](https://github.com/jfhbrook/node-ecstatic/pull/143) lands.
I need some help though. I can't figure out how to test the feature with vows. If I start http-server, I can request files with custom mime-types but no matter what I do in [http-server-test.js](https://github.com/dotnetCarpenter/http-server/blob/ecstatic/test/http-server-test.js#L169), I get an ECONNREFUSED. I'm pretty sure it's a race condition but I can't figure out how to fix it.

Any ideas?

Issue #35, opened 2013-06-20
Configure additional mime types


olegomon:
add mime type support to enable streaming of video files with html 5 video tag

dotnetCarpenter:
I'll give it shot. Just got an API implemented in the [upstream ecstatic](https://github.com/jfhbrook/node-ecstatic/pull/143/). If it gets accepted and I'm successful here... well, then it's done.

dotnetCarpenter:
Can someone give me a hand with #152?
I can't figure out how to test the feature with vows. If I start http-server, I can request files with custom mime-types but no matter what I do in [http-server-test.js](https://github.com/indexzero/http-server/pull/152/files#diff-7a5fd917883091db4c5e5471a21bfb85R165), I get an ECONNREFUSED. I'm pretty sure it's a race condition but I can't figure out how to fix it.

Any ideas?

timdp:
It'd also be nice if we could map `.mpd` to `application/dash+xml` for MPEG-DASH.

stowball:
I would love to be able to serve SVGs correctly using the MIME type `images/svg+xml`

BigBlueHat:
#152 just needs tests afaict. If anyone wants to right those, I'd be 👍 for merging that. Setting media types correctly is rather essential--and `ecstatic` already supports a way to set the mappings (between extension and media type) properly. So. Who's up for writing tests?! 😁

75lb:
MIME types [can be customised in local-web-server](https://github.com/lwsjs/local-web-server/wiki/How-to-customise-response-MIME-types), is that the kind of behaviour you're looking for?

grokify:
+1 for SVG and/or customizable approach similar local-web-server.

disfeng:
+1

anderspitman:
Just want to chime in that lack of WebAssembly mime is forcing me to look for an alternative local server. http-server has beautifully handled all my needs up to this point and I'd hate to switch if I don't have to.

paulmelnikow:
@anderspitman Did you find another option? At some point I switched from `http-server` to `serve`, but it has the same limitation.

anderspitman:
@paulmelnikow I've been using https://crates.io/crates/https. That page includes installation instructions for non Rust developers as well.

yiminghe:
https://crates.io/crates/https: http -m wasm:application/wasm

ddennedy:
[...]

PR #463, opened 2018-09-03
Add support for brotli encoding

thornjad:
As requested in #445, this adds support for serving brotli-encoded files by passing `--brotli` or `-b` to http-server. This necessitates bumping up the minimum `ecstatic` dependency version to `3.3.0`, but this [shouldn't have any breaking changes](https://github.com/jfhbrook/node-ecstatic/blob/master/CHANGELOG.md).


filipesilva:
@indexzero @BigBlueHat anything I can help out with to get this merged?

I tested it locally by following these steps:
- clone this repo
- copy `package-lock.json` to `public/package-lock.json`
- encode `public/package-lock.json` (78kb) to `public/package-lock.json.br` (17kb) via https://dominikhlbg.github.io/brotlijs/
- run `./bin/http-server`
- open chrome at http://localhost:8080/package-lock.json, open devtools-network, ctrl+f5, see original package-lock size
- terminate previous process, run `./bin/http-server -b`
- open chrome at http://localhost:8080/package-lock.json, open devtools-network, refresh, see reduced package-lock size

I think this PR is working as intended 👍

filipesilva:
Heya, don't mean to be a bother but just wanted to ping about this again since there's a project I'm working on that's waiting for this support. Keep up the good work!

thornjad:
@BigBlueHat do you have the ability to merge? And if so, what do you think?

BigBlueHat:
@thornjad sorry I hadn't responded earlier other than the labeling.

Since this bumps the underlying `node-ecstatic` to v3.3.0, I want to be sure that doesn't have any unexpected consequences. IIRC there was some significant shift in the MIME libraries in use which may effect custom type loading for some folks.

We've also got a bit of a tangle with #464, but since you're the author of both, maybe you can help queue those in your preferred order? 😃 

Regardless, I do want to get this merged--and ideally soon.

We should also get some tests written for the `.br` and the priority order thing around `.br` priority vs. `.gzip`.

thornjad:
Hey, that's alright. 

Tests are a good idea. I did put some in `ecstatic`, so we are partially covered that way, but `http-server` should be properly testing itself. I actually keep a sticky note on my desk that just reads "You need more tests".

For the MIME libraries, tests would probably be good here as well, but I've never used custom types in real life, so I'm not sure what to look for.

As for #464, either PR can go first and the other can be updated to match. This PR is a more meaningful change, so maybe this one should go first.

BigBlueHat:
@thornjad wrt to mime stuff, I may be remembering this all wrong.  It looks like #35 is a pending request which has an old PR #152 which may be obviated by `node-ecstatic`'s direct [mime type customization](https://github.com/jfhbrook/node-ecstatic#optsmimetypes).

PR #472 exposes the raw JSON mime config to the command line, but doesn't handle the `.types` loading which `node-ecstatic` makes available as a path set on `--mime-types`.

In short, this is probably not a blocker. 😄 But it is something there's long standing desire for, so maybe we tackle that once much of this is in place.

With some basic tests for this code, I'd be say it's ready to go (it reads well anyway).

Thanks for the help @thornjad!

wentjun:
Hi! I have been following this issue for a while. May I know when will this be merged/released?

thornjad:
@wentjun sorry for the delay! I am working on a giant merge train so this should be merged soon!

filipesilva:
Awesome, happy to see this go in!

Issue #450, opened 2018-06-15
Make Boolean command line switches case insensitive


BigBlueHat:
See #200. Specifically https://github.com/indexzero/http-server/issues/200#issuecomment-347841078

iamsdas:
I'd like take up this issue

thornjad:
On closer inspection, this was actually fixed by #622

Issue #200, opened 2015-10-25
How to disable directory listings?


sapegin:
Readme just says how to enable it but it’s already enabled by default.

frjalex:
Same question here

frjalex:
the 'False' param just wont work

sebgymn:
`-d false`

lunyx:
I'm having this same issue as well.  

Running: http-server /www -d False -i False -p 80

The directory listing still shows up.

kamakazikamikaze:
This issue can be closed.

@bekce Submitted the correct answer. The reason why your directory listings still show is because of capitalization: `False` is a string, `false` is a boolean

lunyx:
That explains a lot... The readme has `-d Show directory listings (defaults to 'True')`, which implied to me that `False` should've also started capitalized.

kamakazikamikaze:
I didn't think to change the capitalization until I invoked `-h` in the CLI which displays it as lowercase. At the moment I was actually working on reading in JSONs in Python where the capitalization is also different (`true` --> `True`), so it suddenly made sense.

One of those "subtle but significant" factors that make a world of difference when auto-casting variable types...

frjalex:
I think it's not an urgent issue, as it's always far more convenient to use the server as a nodejs module but not as-an application.

killroy42:
Did this change recently? I've been using http-server for a long time as a quick way to point port 80 at my various test servers. Suddenly I got listings which I never got before. Took me a while to find this thread... PS: on the command line everything is a string. It might make more sense to either accept all capitalizations, or to output an error for an "invalid value for d", for example.

BigBlueHat:
Closing in favor of #450. I'll also update the README (as it uses init-caps...confusingly).

BigBlueHat:
See 94329688b5bf6c058a9f2c24739bd628b7345dce

PR #464, opened 2018-09-04
Renovate CLI switch parsing

thornjad:
This commit makes a handful of closely-related changes, all aimed at refactoring `bin/http-server` for better maintainability.

- Switched from deprecated [`optimist`](https://www.npmjs.com/package/optimist) to the smaller [`minimist`](https://www.npmjs.com/package/minimist) for CLI parsing
- Used aliases to link together switches with both a `-` and `--` option, thereby removing the need for patterns like `argv.g || argv.gzip`
- Set defaults on almost all switches. This can make debugging easier and makes default values more clear
- Reformatted the help dialog
- Clarified the help dialog to take a sidestep around #450. As noted in https://github.com/indexzero/http-server/issues/200#issuecomment-241065301, `False` is a string, so it should not work on a boolean switch. Instead, this new help dialog encourages using the `--no` prefix. For example, `-i` enables autoIndex, and `--no-i` disables it. This is already possible, so it's not actually an interface change
- As far as I can tell, this PR introduces no user-facing changes except for the help dialog


bin/http-server
```diff
@@ -18,41 +56,40 @@ if (argv.h || argv.help) {
...
-    '  -g --gzip    Serve gzip files when possible [false]',
-    '  -e --ext     Default file extension if none supplied [none]',
-    '  -s --silent  Suppress log messages from output',
-    '  --cors[=headers]   Enable CORS via the "Access-Control-Allow-Origin" header',
-    '                     Optionally provide CORS headers list separated by commas',
-    '  -o [path]    Open browser window after starting the server',
-    '  -c           Cache time (max-age) in seconds [3600], e.g. -c10 for 10 seconds.',
-    '               To disable caching, use -c-1.',
-    '  -U --utc     Use UTC time format in log messages.',
+    '  -p\t           Port to use [Default: 8080]',
```
BigBlueHat: What are the `\t`'s for throughout this section?

bin/http-server
```diff
@@ -18,41 +56,40 @@ if (argv.h || argv.help) {
...
     '',
-    '  -P --proxy   Fallback proxy if the request cannot be resolved. e.g.: http://someurl.com',
+    '  -P --proxy\t   Fallback proxy if the request cannot be resolved. e.g.: http://someurl.com',
     '',
-    '  -S --ssl     Enable https.',
-    '  -C --cert    Path to ssl cert file (default: cert.pem).',
-    '  -K --key     Path to ssl key file (default: key.pem).',
+    '  -S --ssl\t   Enable https [Default: disabled]',
+    '  -C --cert\t   Path to ssl cert file [Default: cert.pem]',
+    '  -K --key\t   Path to ssl key file [Default: key.pem]',
```
BigBlueHat: Thanks for making these match the others!

bin/http-server
```diff
@@ -7,9 +7,47 @@ var colors     = require('colors/safe'),
     httpServer = require('../lib/http-server'),
     portfinder = require('portfinder'),
     opener     = require('opener'),
-    argv       = require('optimist')
-      .boolean('cors')
-      .argv;
+    parseArgs  = require('minimist');
+
+var argv = parseArgs(process.argv.slice(2), {
+  alias: {
```
BigBlueHat: Can we switch the aliases, so the variable names are clearer? `argv.silent` instead of `argv.s`?

bin/http-server
thornjad: Those are tabs to make things line up better. In hind sight, this would be better handled by another library, or just not handled at all

bin/http-server
thornjad: 👍 Makes it more clear, I think

bin/http-server
thornjad: Sure! That's a good idea. I'll make those changes in a bit

bin/http-server
BigBlueHat: "This" being the alignment? I've done some work with http://yargs.js.org/ and really enjoyed it. Certainly less minimal...and would probably need full-on code restructuring...but if you're looking for something like this generally, yargs seems pretty fabulous.

thornjad: I did actually start out replacing optimist with yargs, but I went with minimist instead just because its so much smaller (minimist has no dependencies)

thornjad: I just replaced the tabs with spaces. I really think spaces are the way to go, I don't know what I was thinking before.

thornjad:
I am officially laying this PR down in favor of nconf in #131, it's very well done over there.

Issue #466, opened 2018-09-18
support --port


johnstonmatt:
`-p` flag works, but long form --port should work too. Its a non-breaking change, trivially easy, and makes things more declaritive. I can make a PR if @indexzero gives it a green light?

BigBlueHat:
Send along that PR @johnstonmatt. Thanks!

johnstonmatt:
hey all, this PR failed on travis, but I believe that was not a result of the change. Is this good to be merged?

NightMachinary:
I installed the latest version just now, and --port still doesn't work and doesn't produce an error message either.

thornjad:
@NightMachinary what was the exact command you used and what was the output?

webdesignberlin:
Usage of
`npx http-server ./dist --port 5454`

start server but only with default port 8080:
`npx: Installierte 26 in 2.235s
Starting up http-server, serving ./dist
Available on:
  http://127.0.0.1:8080
  http://10.100.179.79:8080
Hit CTRL-C to stop the server`

PR #468, opened 2018-09-23
johnstonmatt/--port

johnstonmatt:
I added the ability to use a long form of port, `--port` as it is backwards compatible, makes the api more discoverable and declaritive. See #466.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ This will install `http-server` globally so that it may be run from the command 
 
 ## Available Options:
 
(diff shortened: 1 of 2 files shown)


johnstonmatt:
the Travis check failed, but I'm quite certain that is not because of my change

BigBlueHat:
@johnstonmatt correct. It was something breaking at Travis and unrelated to your code. 😄 I've just rerun it and it ran great.

Thanks for this! Merging shortly.

PR #470, opened 2018-10-05
feat(lib): Add the content encoding entity header gzip

NastyZ98:
This feature add the following:

- Addition of content-encoding "gzip" when a compressed file is requested

### DIFF
--- a/lib/http-server.js
+++ b/lib/http-server.js
@@ -57,7 +57,16 @@ function HttpServer(options) {
   }
 
   var before = options.before ? options.before.slice() : [];
+  if (this.gzip) {
+    before.push(function (req, res) {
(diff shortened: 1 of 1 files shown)


msidhoum:
+1

thornjad:
@NastyZ98 this should already be happening, have you experienced the Content-Encoding not being set correctly?

thornjad:
Would you mind trying out the same thing using the current `master` branch? If it still isn't working, could you comment with the errors you're seeing and a list (or screenshot) of the headers that _are_ being sent? This may actually be a bug in `ecstatic`

thornjad:
Closed since the branch for this PR has been deleted

PR #472, opened 2018-10-23
Support custom mime types (content-type)

599316527:
Just pass the mimeTypes option to node-ecstatic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,8 @@ This will install `http-server` globally so that it may be run from the command 
 
 `-e` or `--ext` Default file extension if none supplied (defaults to `html`)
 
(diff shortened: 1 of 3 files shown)


thornjad:
👍 lgtm

Smilebags:
Why has this not been merged? Such a simple change which solves so many problems. I'll be hand-copying these changes so I can use http-server to serve wasm modules.

nickolay:
Even with this patch I wasn't able to use wasm modules -- probably due to https://github.com/indexzero/http-server/issues/480 . As noted there, with ecstatic 4.0.0 wasm was served with the correct content-type out-of-the-box.

J-Rojas:
@nickolay #565 solves the wasm problem along with supporting mime types.

thornjad:
Closing in favor of #693, since we'll be able to support it directly

PR #474, opened 2018-10-24
add middleware option

deju:
Anyone just want to serve index.html after webpack build, the router was management by front end developer. So I add middleware option, and write `returnIndex.js` in `middleware` directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,8 @@ This will install `http-server` globally so that it may be run from the command 
 
 `-r` or `--robots` Provide a /robots.txt (whose content defaults to `User-agent: *\nDisallow: /`)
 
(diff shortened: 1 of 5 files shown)


BigBlueHat:
@deju I love the idea of this PR! Thank you.

However, we're needing to refactor http-server soon to remove our dependency on `flatiron/union`, so I'd rather not introduce a feature which uniquely depends on it--see #483.

Once #483 is solved, this seems like a great feature to consider for easier extensibility.

Thoughts welcome!

thornjad:
While I think this is a neat feature, I'm not so sure this is within the scope of http-server, which is a static content server. Would this be better suited as a companion package which depends on http-server, or a middleware-focused library like express?

thornjad:
Closing due to the lack of activity since opening, please discuss and reopen if this feature is still wanted

PR #477, opened 2018-11-07
Handle proxy errors

nLight:
`http-proxy` does not handle errors by default and would just throw crashing the server. This PR adds error handling with rudimental logging.

### DIFF
--- a/lib/http-server.js
+++ b/lib/http-server.js
@@ -111,6 +111,13 @@ function HttpServer(options) {
       proxy.web(req, res, {
         target: options.proxy,
         changeOrigin: true
+      }, function (err, req, res, target) {
+        if (options.logFn) {
+          options.logFn(req, res, {
+            message: err.message,
+            status: res.statusCode });
+        }
+        res.emit('next');
       });
     });
   }


nLight:
I couldn't really figure out how to test this with the testing setup. Any hint would be appreciated.

nLight:
@BigBlueHat anything I could do to move that one forward?

patrikbeno:
> I couldn't really figure out how to test this with the testing setup. Any hint would be appreciated.

In my case, `http-server` crashes if the proxy is not available, so I guess you could simply point proxy to the free localhost port and then hit `http-server` with non-existent URL.

It will fail and abort like this:

```
index.js:120
    throw err;
    ^

Error: connect ECONNREFUSED 127.0.0.1:8080
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1113:14)
```

nLight:
@BigBlueHat any chance looking into this?

nLight:
@patrikbeno figuring out a test case is easy, sure. What I would like to get help with is fitting this test in the testing env of the project.

nLight:
@indexzero ping?

lib/http-server.js
```diff
@@ -107,6 +107,14 @@ function HttpServer(options) {
 
   if (typeof options.proxy === 'string') {
     var proxy = httpProxy.createProxyServer({});
+    proxy.on('error', function (err) {
+      if (options.logFn) { // If there's a log function for requests we can log
```
BigBlueHat: If there's a log function set, the code should use that and not just `console.info`. See https://github.com/indexzero/http-server/blob/94329688b5bf6c058a9f2c24739bd628b7345dce/lib/http-server.js#L62-L64

nLight: Should I fake `req` and `res` arguments in this case? I don't have them in this context

nLight: ping

nLight:
Any progress? @BigBlueHat crashing proxies is not cool

lib/http-server.js
BigBlueHat: Move the `on('error')` code into the `before.push` https://github.com/indexzero/http-server/blob/f65f1a7154456f823261c4f8e90b35eff66eb23a/lib/http-server.js#L118 where the proxy is added into the mix.

nLight: Found a cleaner way to do it. Force-pushed, please review 🙇

nLight:
Just a friendly reminder

nLight:
Just another friendly reminder @BigBlueHat

nLight:
Thank you 🙇

nLight:
🎉

PR #482, opened 2018-11-29
feature delay

aemonge:
Implemented a delay for the response, for better usage as a mock server :)


BigBlueHat:
@aemonge since folks also use http-server in production, I'm not sure about this as a core switch...

Perhaps simply writing a wrapper around http-server that injects the delay?

aemonge:
I don't understand your fear/worries. 

Regardless, you mean a wrapper around it as in refactor the code so It can be injected on demand, or to have an extension to the library that implements it ?

BigBlueHat:
@aemonge it's mostly a question of scope creep for the project. The code you've got here is essentially additional middleware. What's possibly needed is a way for more easily configurable middleware in addition to the core static hosting features of `http-server`.

Right now, the easiest option would be to build a small wrapper script (maybe `slow-http-server` 😉) that implements this middleware and it's own delay parameter.

For the time being, though, I don't see this sort of feature being on we'll pursue.

@thornjad thoughts?

thornjad:
I agree, I think this is out of scope for `http-server`, but would probably be a useful tool. `slow-http-server` (or whatever you want) could intercept requests and delay before passing them to `http-server` (and/or intercept the server response), maybe even having settings for simulating different network speeds, and other tools.

aemonge:
Thanks both, keep me in touch if you develop it. And I'll do the same ;)

Issue #483, opened 2018-12-04
Remove flatiron/union dependency


BigBlueHat:
We need to get off the outdated flatiron/union dependency because it's no longer necessary because Streams--see https://github.com/flatiron/union/pull/61#issuecomment-443856321 for more.

@thornjad this relates to the PR you sent to flatiron/union. If you've still got coding cycle time, I'd love your help on this one!

PR's from anyone interested are also welcome. 😁

thornjad:
@indexzero @BigBlueHat does union only provide middleware capability and buffering? Or is it more than that?

BigBlueHat:
@indexzero would be the one to answer the `union` question. Regardless, getting up to something more maintained is important to fix several related issues.

Thanks for all you're doing @thornjad!

indexzero:
@thornjad primarily – `union` was created when there were no buffered streams natively in Node.js to address that shortcoming. That obviously has not been the case for many years.

In addition to native buffering it's also the main entry point for how the http(s) server is created. That aspect of `union` has been diligently maintained and improved on over the last several years as [`create-servers`](https://github.com/indexzero/create-servers).

With that in mind, we would just need a small "stack" library like `connect`. Then [this bit](https://github.com/indexzero/http-server/blob/master/lib/http-server.js#L118-L139) of code could become:

``` js
// 
// ... rest of the existing code above ...
// ... replacing res.emit('next') by invoking a next parameter ...
// 
  if (this.headers) { 
    before.unshift((req, res, next) => {
      Object.keys(this.headers).forEach(header => {
        res.setHeader(header, this.headers[header]);
      });

      next();
    });
  }

  before.push(function onError(err, req, res, next) {
    if (options.logFn) {
      options.logFn(req, res, err);
    }

    res.end();
  });

  const handler = connect();
  before.forEach(layer => handler.use(layer));

  this.serverOptions = { handler };
  if (options.https) this.https = options.https;
}

HttpServer.prototype.listen = function (port, host, callback) {
  const key = this.serverOptions.https ? 'https' : 'http';

  if (!callback) {
    if (typeof host === 'function') {
      callback = host;
      host = null;      
    } else {
      callback = function () {};
    }
  }

  this.serverOptions[key].port = port;
  if (host) this.serverOptions[key].host = host;

  createServers(this.serverOptions, (err, servers) => {
    if (err) return callback(err);
    this.server = servers[key];
  });
};
```

[...]

PR #485, opened 2018-12-24
Updated dependencies to fix security issues

chill117:
Removed common-style (deprecated, depends on unpatched packages)

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,19 +66,19 @@
     }
   ],
   "dependencies": {
-    "colors": "1.0.3",
-    "corser": "~2.0.0",
-    "ecstatic": "^3.0.0",
-    "http-proxy": "^1.8.1",
-    "opener": "~1.4.0",
-    "optimist": "0.6.x",
-    "portfinder": "^1.0.13",
-    "union": "~0.4.3"
+    "colors": "^1.3.3",
+    "corser": "^2.0.1",
+    "ecstatic": "^3.3.0",
+    "http-proxy": "^1.17.0",
+    "opener": "^1.5.1",
+    "optimist": "~0.6.1",
+    "portfinder": "^1.0.20",
+    "union": "~0.5.0"
   },
   "devDependencies": {
     "common-style": "^3.0.0",
-    "request": "2.49.x",
-    "vows": "0.7.x"
+    "request": "^2.88.0",
+    "vows": "~0.7.0"
   },
   "bugs": {
     "url": "https://github.com/nodeapps/http-server/issues"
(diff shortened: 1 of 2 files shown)


BigBlueHat:
Thanks for sending these! We've actually got quite a tangle of dependencies that need upgrading (or removing/reconsidering).

@thornjad not sure if this is the "tip of the spear" in that regard or not, but certainly bumping all the versions is important before the next major (and possibly next minor--if we can avoid BC breaks) release.

thornjad:
I certainly agree this is important. However I do think we need to replace `common-style` rather than just remove it. I think `eslint` is a good contender, though there are several other good options as well.

package.json
```diff
@@ -66,18 +65,19 @@
...
-    "opener": "~1.4.0",
-    "optimist": "0.6.x",
-    "portfinder": "^1.0.13",
-    "union": "~0.4.3"
+    "colors": "1.3.3",
+    "corser": "2.0.1",
+    "ecstatic": "3.3.0",
+    "http-proxy": "1.17.0",
+    "lodash": "4.17.11",
+    "minimatch": "3.0.4",
```
thornjad: While I love lodash, why has it and minimatch been added here?

package.json
```diff
@@ -13,7 +13,6 @@
   ],
   "scripts": {
     "start": "node ./bin/http-server",
-    "pretest": "common bin/http-server lib/ test",
```
thornjad: Style checking as part of testing is very important; that way the code is consistently readable. So while `common-style` is deprecated, I think it should be replaced rather than removed.

lvl99:
Apparently the version of `qs` used in `http-server` has some potential issues: https://app.snyk.io/test/npm/http-server/0.11.1

package.json
thornjad: I've made PR #492 for discussion with regards to `common-style` and friends.

thornjad:
@lvl99 you're right, and it's already fixed in the next version of `union`, so updating dependencies should fix that. It's because of that vulnerability that this is high on my radar. 

@chill117 would you mind changing this to use version ranges instead of specific versions? You can also mark this as addressing #483

chill117:
I've updated the version numbering to include ~ and ^ where appropriate. I returned common and the pretest for code-style checking - to limit the scope of this PR to updating non-dev dependencies.

BigBlueHat:
@thornjad there's a suggestion for prettier as a replacement to common-style in #494. Certainly this can wait, but wanted to connect dots.

This looks good to merge (pending your merge-train 🚆). 😄

thornjad:
Choo choo!

I also made a suggestion of eslint in #497, but I meant that more as a discussion (and didn't see #494 come in before it). And of course draft PRs become a thing just after I need it.

PR #488, opened 2019-01-09
Point out the npx option

wmertens:


BigBlueHat:
Love it! Thanks @wmertens!

PR #493, opened 2019-01-27
add super verbose, show unknown income request header and body

foxundermoon:


bin/http-server
```diff
@@ -40,6 +40,7 @@ if (argv.h || argv.help) {
     '',
     '  -r --robots  Respond to /robots.txt [User-agent: *\\nDisallow: /]',
     '  --no-dotfiles  Do not show dotfiles',
+    '  -V --superverbose    super verbose mode',
```
thornjad: Could we move this right next to `--silent` above, since they're both about logging?

bin/http-server
```diff
@@ -80,6 +82,54 @@ else if (colors) {
     request: function () {}
   };
 }
+if (superverbose) {
```
thornjad: Since this is an expanded version of the already-existing `logger.request`, could we move this inside the `if` block on line 57, rather than define a function and then immediately override it?

bin/http-server
```diff
@@ -80,6 +82,54 @@ else if (colors) {
...
+      if (error) {
+        logger.info(
+          '[%s] "%s %s" Error (%s): "%s"',
+          date, colors.red(req.method), colors.red(req.url),
+          colors.red(error.status.toString()), colors.red(error.message)
+        );
+        logger.info('---------------header--------------');
+        var  request = req.request;
+        var rawHeaders = request.rawHeaders;
+        if (rawHeaders && rawHeaders.length > 0) {
```
thornjad: Since `0` is falsy, we can just check 
```suggestion
        if (rawHeaders && rawHeaders.length) {
```

bin/http-server
```diff
@@ -80,6 +82,54 @@ else if (colors) {
...
+          var headerStr = '';
+          for (var i = 0; i < rawHeaders.length;i++) {
```
thornjad: Rather than a `for` loop, it might be cleaner and easier to use `rawHeaders.reduce` here. What do you think?

bin/http-server
```diff
@@ -80,6 +82,54 @@ else if (colors) {
...
+          req.request.on('data', function (chunk) {
+            body.push(chunk);
+          }).on('end', function () {
+            body = Buffer.concat(body).toString();
+            logger.info('-------------request body----------');
+            logger.info(body) ;
+          });
+        }
+      }
+      else {
```
thornjad: Since this is "superverbose", is there more information we can show if there was no error?

bin/http-server
```diff
@@ -80,6 +82,54 @@ else if (colors) {
...
+    info: console.log,
+    request: function (req, res, error) {
+      var date = utc ? new Date().toUTCString() : new Date();
+      if (error) {
+        logger.info(
+          '[%s] "%s %s" Error (%s): "%s"',
+          date, colors.red(req.method), colors.red(req.url),
+          colors.red(error.status.toString()), colors.red(error.message)
+        );
+        logger.info('---------------header--------------');
```
thornjad:
```suggestion
        logger.info('---------------headers-------------');
```

bin/http-server
```diff
@@ -80,6 +82,54 @@ else if (colors) {
...
+            if (i % 2 === 0) {
+              headerStr += ': ';
+            }
+            else {
+              headerStr += '\n';
+            }
+          }
+          logger.info(colors.cyan(headerStr));
+        }
+        if (req.readable) {
```
thornjad: I'm not a big fan of logging the request body, I think it might be a bit _too_ verbose. What do you think?

bin/http-server
```diff
@@ -80,6 +82,54 @@ else if (colors) {
     request: function () {}
   };
 }
+if (superverbose) {
+  logger = {
+    info: console.log,
+    request: function (req, res, error) {
+      var date = utc ? new Date().toUTCString() : new Date();
+      if (error) {
```
thornjad: Should we be showing all this info if there was an error, or if there was no error? Was the logic accidentally flipped here?

PR #497, opened 2019-02-03
WIP: ES2015+

### LINKED ISSUE #492: Allow ES2015+ syntax
Style linting (and therefore `npm test` and Travis) does not currently allow ES2015, even though Node fully supports it at this point, and an upgrade has a lot of real benefits; the first that comes to mind is `const` and `let`. ES2017 also adds async/await, which I have found extremely useful.

I see two ways we can go forward with this, we can either refactor http-server into ES2015+, or we can simply allow ES2015+ additions (like #491). Refactoring could be nice (I like refactoring), but would probably require a lot of existing PRs to be refactored as well. But this could just be shallow refactoring like switching `var`s to `let` and `const`. Simply allowing ES2015+ to pass tests, on the other hand, would allow it to be added in PRs and everything could be refactored piecemeal.

In either case, we should probably also bump up the versions of Node that Travis tests on, as well as decide on a new linter (common-style is deprecated in favor of eslint-config-populist).

thornjad:
This accomplishes two things:

- First, it replaces the deprecated `common-style` with its successor `eslint-config-populist`. This includes allowing ES2015+ syntax.
- Second, it starts the process of bringing `http-server` into line with `populist` style.

Now, @BigBlueHat @indexzero I've run into a few issues that I'd like some input on -- there are some problems with using `populist` style:

- There is some issue in detecting and allowing indentation to keep declarations in line. For example, this throws the linting error `Expected indentation of 2 spaces but found 4`, even though this is allowed according to the printed rules:

```
var colors = require('colors'),
    os     = require('os');
```

- I disagree with the rule about using the `async` package over native Promises, especially now that `async`/`await` is widely implemented.
- There are a number of linting errors that would take more significant changes. For example, there is an `lstatSync` call that should be made asynchronous, and there are complexity warnings. I haven't tackled these here.

For the moment, I've disabled linting rules in select errors, but _this is bad_ (citation needed). We should have a discussion about where to go from here.

Fixes #492

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
 node_modules/*
 npm-debug.log
-.dir-locals.log
+.dir-locals.el
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules/*
 npm-debug.log
+.dir-locals.log
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,9 @@
 sudo: false
 language: node_js
 node_js:
-  - "4"
-  - "5"
-  - "6"
-  - "7"
+  - "8"
+  - "10"
+  - "11"
 
 before_install:
   - travis_retry npm install
--- a/bin/http-server
+++ b/bin/http-server
@@ -1,15 +1,13 @@
 #!/usr/bin/env node
 
-'use strict';
-
 var colors     = require('colors/safe'),
     os         = require('os'),
     httpServer = require('../lib/http-server'),
     portfinder = require('portfinder'),
     opener     = require('opener'),
-    argv       = require('optimist')
-      .boolean('cors')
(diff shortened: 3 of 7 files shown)


BigBlueHat:
Given that @indexzero wrote the populist tooling, we should probably await (hehe) input (especially wrt to asynx). 😉 

> For the moment, I've disabled linting rules in select errors, but _this is bad_ (citation needed). We should have a discussion about where to go from here.

While I agree with this generally, I would add that if this change moves things to more stable tooling ground while still maintaining current code styling (even with a few things skipped/ignored), I'm OK with moving forward.

@indexzero we await your sage response. 😺

thornjad:
This has fallen behind master by so much I'm just going to close. This is something we can work in bit by bit if we want

PR #498, opened 2019-02-03
add stale bot config

thornjad:
This would configure [stale bot](https://github.com/apps/stale), which would then be added to this repository. This would help cut down on the number of issues and PRs we have open right now.

After acceptance and merging, I can add the bot integration.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,27 @@
+# probot-stale configuration -- closes issues and PRs that have gone stale
+
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 90
+
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 10
+
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+
+# Label to use when marking an issue as stale
+staleLabel: stale
+
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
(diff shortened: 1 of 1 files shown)


thornjad:
@BigBlueHat what do you think of this? I would like your (or @indexzero's) go-ahead before adding an integration.

BigBlueHat:
My one concern is that we'd actually frustrate a lot of people when the bot goes through and closes a bunch of old (but still useful) issues...

We do need to (continue to) do house keeping here, but most of the things that are stale here are stale because our team is small, not because there's not interest, activity, or even PRs...per se. Just lack of time to review, merge, and mix all the codez.

Happy to hear the counter-case, though. 😃

thornjad:
That's my chief concern as well, but I think stale issues/PRs are bad as well. The staleness can easily be staved off by the submitter as well, however. I went with the default message here, but it could be changed to something friendlier, and with a prompt to reply with some message to add activity. The timeframe can also be extended, maybe 120 days? I think its reasonable if no one, including the submitter says or does anything for 120 days it can be considered stale enough to close.

For current issues that are over this limit, there could be some special label to add to them which adds activity and grandfathers them into the system. Or the integration could just wait until we've gone through everything real quick.

On the other hand, maybe this is really just not necessary? When I have the time to go through a bunch of issues I've been sorting by activity ascending so I'll hopefully eventually make it through all the "stale" stuff anyway...

thornjad:
Closing in favor of just trying to keep up with everything. We may want to adopt a policy of closing PRs which have requested changes but no activity in some large amount of time.

PR #499, opened 2019-02-03
Update license year

thornjad:
I have no idea if the copyright-holder list is accurate.


BigBlueHat:
👍 I think leaving the copyright alone for now is fine. "the Contributors" essentially means (afaik) anyone who's put code into this--which means its a "mixed" copyright owner.

BigBlueHat:
Relatedly, we should make sure folks get their names listed in package.json for meaningful contributions (i.e. something greater than a few small lines or a package.json dependency version bump).

PR #504, opened 2019-02-07
add --overlay option

dashie:
Add the possibility to specify an optional --overlay folder, if the file to serve exists in this folder this file overlay the base file in the root folder

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,8 @@ Using `npx` you can run the script without installing it first:
 
 `-r` or `--robots` Provide a /robots.txt (whose content defaults to `User-agent: *\nDisallow: /`)
 
+`--overlay` Add an overlay folder. Any requested file is previously searched in this folder and if does not exists then is searched in the default root folder.
+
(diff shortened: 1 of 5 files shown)


BigBlueHat:
What's the use case?

thornjad:
Would you be able to do this by serving the "overlay" directory and using the main directory as a proxy? Then the overlay directory would be checked first, and if it doesn't have a file, it goes the the proxy.

dashie:
> Would you be able to do this by serving the "overlay" directory and using the main directory as a proxy? Then the overlay directory would be checked first, and if it doesn't have a file, it goes the the proxy.

Yes I know, but I don't want to run 2 processes and bind 2 ports.
But the behavior would be the same.

thornjad:
I think this is out-of-scope for http-server, since the proxy option would have the same effect. Without a proxy, if a file isn't found, it's a 404 error.

This would be a good use case for a small independent package which uses the proxy method automatically, given a root and overlay directory.

Issue #532, opened 2019-05-23
Silent but shown initial output


Nogbit:
I would like to request a feature.  I know #381 is closed (asked before) however my ask is different.

When automating via npm scripts the running of a server and the extra args to http-server the initial output is very handy.  It shows what directory is being served as well as providing a clickable url in the output of the different urls and ports it's being served on.

So the ask is, either another arg that is `--show-start` or to change `--silent` to show at a minimum the initial output and nothing else.  Or something similar that provides that output while also silencing the request logs.

Xmader:
#538

thornjad:
The startup output was expanded slightly in v13.0.1, if you use that version, does the `--silent` option now give info you're looking for? If not, could you give an example of exactly the extra output you'd like?

dmnsgn:
The PR above with `--nv --no-verbose Only show the basic info',` option has not been merged so the issue is still valid.

PR #538, opened 2019-06-05
add a option of no verbose

Xmader:

Add a option to prevent showing all the request info, but show the basic info.

#532

#381

Add a option of no verbose

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,8 @@ Using `npx` you can run the script without installing it first:
 
 `-s` or `--silent` Suppress log messages from output
 
+`--nv` or `--no-verbose` Only show the basic info
+
 `--cors` Enable CORS via the `Access-Control-Allow-Origin` header
 
 `-o [path]` Open browser window after starting the server. Optionally provide a URL path to open. e.g.: -o /other/dir/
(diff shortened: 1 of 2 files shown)


thornjad:
I'm not sure of the usefulness of this option. The info is already shown by default, and will be the only output before any requests are made (providing the clickable link in some terminal emulators, as mentioned in the linked issues). And then there is already `--silent` to suppress all logging. What is the additional benefit of this "sort-of-silent" option?

developit:
The use-case is this: spawning a dev server for folks in scaffolded projects, many developers don't know what port `http-server` was spawned on (especially if it auto-selected a free port). That makes the URL output immensely valuable, since otherwise you have to know that the default is 8080 and it uses an incrementing search to find free ports.

However, verbose request logging for a development server is not workable: it interlaces request logs with far more important information, like build information and errors. There are many cases where request logging is unnecessary, yet where the display of a server's listen() address is still important.

thornjad:
For use in scaffolding, would it be sufficient to have the scaffold pass a predetermined port into http-server and tell the user that port?

PR #546, opened 2019-07-19
Add support for Access-Control-Expose-Headers using 'cors' configurat…

### LINKED ISSUE #545: Add support for 'Access-Control-Expose-Headers' using --cors
When using XHR requests, only certain default headers are approved for access. The 'Access-Control-Expose-Headers' cors header allows the XHR requests to read additional headers outside of this default list. It would be helpful to add support for this by adding the entire list of cors headers supplied in the configuration parameter to this response header.

https://github.com/mozilla/pdf.js/issues/3150#issuecomment-17582371
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers

J-Rojas:
Fix #545. Supply 'corser' with a similar set of responseHeaders from the 'cors' options.

### DIFF
--- a/lib/http-server.js
+++ b/lib/http-server.js
@@ -99,12 +99,17 @@ function HttpServer(options) {
   if (options.cors) {
     this.headers['Access-Control-Allow-Origin'] = '*';
     this.headers['Access-Control-Allow-Headers'] = 'Origin, X-Requested-With, Content-Type, Accept, Range';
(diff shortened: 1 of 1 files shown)


thornjad (changes requested):
I like the way this looks! Only two items left to do:

- [ ] Merge `master` again, there are new tests to run on this PR
- [ ] Add this header to the CORS test in `test/main.test.js`

Issue #431, opened 2018-02-19
Automatically create self-signed cert?


laggingreflex:
Would it be a good idea to create self-signed certificate to be used when enabling https? Using something like [selfsigned](https://github.com/jfromaniello/selfsigned) or [https-pem](https://github.com/watson/https-pem)

BigBlueHat:
#402 is similar and references how to make one using `openssl` on the command line.

Do you know of any self signed cert tools that work in a similar command line fashion as openssl?

I'd be 👍 on including one as a dev dependency, but I'm not keen to make managing cert creation part of this project (trying to avoid scope creep).

AJLoveChina:
For automatically  self-signed cert
you can try this : https://github.com/AJLoveChina/http-server-v2

@laggingreflex

PR #562, opened 2019-09-20
feat: auto gen self signed cert

### LINKED ISSUE #402: Make https DX better
Hey,
first of all great project! 

I would recommend one improvement. It would be nice if `http-server -S path` would generate automatically some dummy ssl certs for me so i don't need to do it by myself. `webpack-dev-server` does it and it's great. Otherwise I need to generate them with something like (this at least should be part of the readme):

`openssl req -newkey rsa:2048 -new -nodes -x509 -days 3650 -keyout key.pem -out cert.pem`

Best!

wood1986:

automatically generate a self-signed certificate. fix #402 and #431

I only make the change to bin/http-server

I do not know how to add test to the bin/http-server

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 node_modules/*
 npm-debug.log
 .dir-locals.el
+bin/pem
--- a/bin/http-server
+++ b/bin/http-server
@@ -7,6 +7,8 @@ var colors     = require('colors/safe'),
     httpServer = require('../lib/http-server'),
     portfinder = require('portfinder'),
     opener     = require('opener'),
+    fs         = require('fs'),
+    selfsigned = require('selfsigned'),
     argv       = require('optimist')
       .boolean('cors')
       .boolean('log-ip')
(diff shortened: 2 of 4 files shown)


wood1986:
I hope you guys can take my PR. Thanks

thornjad:
I'm very hesitant about automatically creating certificates, that shouldn't really be the responsibility of a server.

Also, it looks like this PR makes http-server always create a certificate if there isn't one already. What if a user doesn't want to generate a certificate?

wood1986:
>Also, it looks like this PR makes http-server always create a certificate if there isn't one already. What if a user doesn't want to generate a certificate?

https://github.com/wood1986/http-server/blob/7b670ed24d0b48bdd8cd05d4a824e174e3292025/bin/http-server#L140

has already respected that. If people passed a cert and it is valid, I will auto generate it. If they do not want to generate it and use their own cert, L140 will not go in

thornjad:
Unless @BigBlueHat or @indexzero disagree with me, I'm going to say this is out-of-scope, i.e. not something http-server should be doing.

AJLoveChina:
@wood1986  
you can try this :  https://github.com/AJLoveChina/http-server-v2

BigBlueHat:
@AJLoveChina thanks for building that for folks who want it. However, could you rename the project to something that doesn't imply that it's part of this project? Thanks!

AJLoveChina:
> @AJLoveChina thanks for building that for folks who want it. However, could you rename the project to something that doesn't imply that it's part of this project? Thanks!

wlc and that's fine, new name is : https://github.com/AJLoveChina/hs2

BigBlueHat:
Thanks, @AJLoveChina!

PR #565, opened 2019-09-25
Support mime-type configuration.

J-Rojas:
Fix for Content-Type: application/wasm

Adds support for mime types using --mime=<JSON string>.

Example:

http-server --mime='{ "text/plain": [ "txt", "md" ] }'

This configuration support Content-Type application/wasm using a hack (only way based on the ecstatic module version used).


madprogramer:
Please merge?

thornjad:
This will be supported directly by #693

PR #569, opened 2019-10-23
do a hacky workaround directly to the http-server package to fix…

### LINKED ISSUE #525: ERR_INVALID_REDIRECT when running http-server
**Do you want to request a *feature* or report a *bug*?**

**If the issue is a bug report, please provide the steps to reproduce it. Please include the actual command causing the issue if applicable.**

npm i -g http-server
http-server

Then open chrome at 127.0.0.1:8080

**What did you expect to happen?**

Open an empty page

**What actually happened? Please include the actual error trace and / or stack trace if applicable.**

ERR_INVALID_REDIRECT

**If the issue is a feature request, what is the motivation / use case for it?**

**Tell us about your environment**
- **http-server version:**
Tried 0.11 and also the master branch from the github repo
- **Platform:**
win10

Xmader:

Fixes #525

do a hacky workaround directly into the `http-server` package
(patch to https://github.com/jfhbrook/node-ecstatic/blob/master/lib/ecstatic.js#L20)

My solution is so hacky. If it has some side effects, please help.
(If there is another function called `decodePathname` also, it will be affected.)

### DIFF
--- a/lib/http-server.js
+++ b/lib/http-server.js
@@ -26,6 +26,7 @@ var _pathNormalize = path.normalize;
 path.normalize = function (p) {
   var caller = getCaller();
   var result = _pathNormalize(p);
+  // https://github.com/jfhbrook/node-ecstatic/blob/master/lib/ecstatic.js#L20
   if (caller === 'decodePathname') {
     result = result.replace(/\\/g, '/');
   }
--- a/lib/http-server.js
+++ b/lib/http-server.js
@@ -6,7 +6,31 @@ var fs = require('fs'),
     auth = require('basic-auth'),
     httpProxy = require('http-proxy'),
     corser = require('corser'),
+    path = require('path'),
     secureCompare = require('secure-compare');
+
+// a hacky and direct workaround to fix https://github.com/http-party/http-server/issues/525
+function getCaller() {
+  try {
+    var stack = new Error().stack;
+    var stackLines = stack.split('\n');
+    var callerStack = stackLines[3];
+    return callerStack.match(/at (.+) \(/)[1];
+  }
+  catch (error) {
+    return '';
+  }
+}
+
+var _pathNormalize = path.normalize;
+path.normalize = function (p) {
+  var caller = getCaller();
+  var result = _pathNormalize(p);
+  if (caller === 'decodePathname') {
+    result = result.replace(/\\/g, '/');
+  }
+  return result;
+};
 
 //
 // Remark: backwards compatibility for previous


thornjad:
How does this hack work? Does this matter on non-NT OS's? Could we make tests to ensure this continues to work?

Xmader:
> Could we make tests to ensure this continues to work?

I think the tests is already covered it.

If this patch is not applied, or not working, the tests would fail with many errors.
- the `When http-server is listening on 8080 when requesting /` test would report `Cannot read property 'statusCode' of undefined`, as well as the tests regards to compression
- the `When http-server is listening on 8080 and options include custom set http-headers` test would report `Cannot read property 'headers' of undefined`
- the `When http-server is listening on 8080 when robots options is activated` test would report `Error: Invalid URI "http:///". This can be caused by a crappy redirection. `, straight forward and pointful
It is the same error that #525 reported.

If this patch is applied, no error reported.

By the way, the test system can't work on Windows unless I upgrade the `vows` package from 0.7.0 to the latest version (0.8.3) .

thornjad:
> writing in progress

Do you have an update? This looks to be hacking into `path` itself, is that right?

Xmader:
yes

Xmader:
On Linux (or other Unix-like operating systems),  `path.normalize("/path//to/file")` returns `"/path/to/file"`
But on Windows, `path.normalize("/path//to/file")` returns `"\path\to\file"`, it is not a valid path component in [URL](https://en.wikipedia.org/wiki/URL).
This patch forces the `path.normalize` function to act like it does on *nix, in this specific call stack.
This doesn't matter on non-NT OS's.

thornjad:
Well really this doesn't look too bad, even if _super_ hacky. Still runs for me on Mac, and tests pass. Well tests fail on Node 12 but I think that's a separate problem. I don't have access to a Windows machine to try this out for sure on there, though.

Xmader:
The problem on Node 12 is related to the `vows` test framework.
When I upgrade `vows` to the latest version (0.8.3), it solves.

thornjad:
Oh duh you said that was an issue! When I upgrade to vows 0.8.3, however, vows seems to hang after all the tests pass, before reporting success. Are you getting the same?

Xmader:
Yes. I just press Ctrl-C to stop it because I see all the tests pass.
It seems that vows 0.8.3 doesn't exit automatically like in 0.7.0 after all the tests pass.
I will try to fix it in a new pull request later today.

thornjad:
@Xmader could you merge master back into this pr? Tests should pass now!

Xmader:
> 
> 
> @Xmader could you merge master back into this pr? Tests should pass now!

no problem

PR #571, opened 2019-11-04
Add TLS/SSL usage description

brizer:

Docs enhancement.

README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,30 @@ http-server --proxy http://localhost:8080?
 
 Note the `?` at the end of the proxy URL. Thanks to [@houston3](https://github.com/houston3) for this clever hack!
 
+## Using SSL
+
+First, make sure that you have `key.pem` and `cert.pem` files. You can generate them using this command:
+
+``` sh
+openssl req -newkey rsa:2048 -new -nodes -x509 -days 3650 -keyout key.pem -out cert.pem
+```
+
+You will be prompted with a few questions after entering the command. Use `127.0.0.1` as value for `Common name` if you want to be able to install the certificate in your OS's root certificate store or browser so that it is trusted.
+
+This generates a cert-key pair and it will be valid for roughly 10 years (3650 days to be exact).
+
+Then you need to run the server with `-S` for enabling SSL and `-C` for your certificate file:
+
+``` sh
+http-server -S -C cert.pem -o
+Starting up http-server, serving ./ through https
+Available on:
+  https:127.0.0.1:8080
+  https:192.168.1.101:8080
+  https:192.168.1.104:8080
+Hit CTRL-C to stop the server
+```
+
 # Development
 
 Checkout this repository locally, then:
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ Note the `?` at the end of the proxy URL. Thanks to [@houston3](https://github.c
 
 ## TLS/SSL
 
-First, make sure that you have `key.pem` and `cert.pem` files. You can generate them using this command:
+First, you need to make sure that [openssl](https://github.com/openssl/openssl) is installed correctly, and you have `key.pem` and `cert.pem` files. You can generate them using this command:
 
 ``` sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -92,6 +92,30 @@ http-server --proxy http://localhost:8080?
 
 Note the `?` at the end of the proxy URL. Thanks to [@houston3](https://github.com/houston3) for this clever hack!
 
+## Using SSL
```
thornjad: To be pedantic, I think we should say "TLS/SSL" here, since the user will technically be enabling TLS, even though people sometimes call it SSL.

README.md
```diff
@@ -92,6 +92,30 @@ http-server --proxy http://localhost:8080?
...
+
+First, make sure that you have `key.pem` and `cert.pem` files. You can generate them using this command:
+
+``` sh
+openssl req -newkey rsa:2048 -new -nodes -x509 -days 3650 -keyout key.pem -out cert.pem
+```
+
+You will be prompted with a few questions after entering the command. Use `127.0.0.1` as value for `Common name` if you want to be able to install the certificate in your OS's root certificate store or browser so that it is trusted.
+
+This generates a cert-key pair and it will be valid for roughly 10 years (3650 days to be exact).
```
thornjad: What do you think of reversing the phrasing here to say

```suggestion
 This generates a cert-key pair and it will be valid for 3650 days (about 10 years).
```

README.md
```diff
@@ -92,6 +92,30 @@ http-server --proxy http://localhost:8080?
...
+
+Then you need to run the server with `-S` for enabling SSL and `-C` for your certificate file:
+
+``` sh
+http-server -S -C cert.pem -o
```
thornjad: `-o` isn't relevant to TLS/SSL, so we shouldn't include it here.

README.md
```diff
@@ -92,6 +92,30 @@ http-server --proxy http://localhost:8080?
...
+Starting up http-server, serving ./ through https
+Available on:
+  https:127.0.0.1:8080
+  https:192.168.1.101:8080
+  https:192.168.1.104:8080
+Hit CTRL-C to stop the server
```
thornjad: Including this with the command may imply that these are also commands, when they are actually output. Maybe we should say "This is what should be output if successful"

README.md
```diff
@@ -92,6 +92,30 @@ http-server --proxy http://localhost:8080?
 
 Note the `?` at the end of the proxy URL. Thanks to [@houston3](https://github.com/houston3) for this clever hack!
 
+## Using SSL
+
+First, make sure that you have `key.pem` and `cert.pem` files. You can generate them using this command:
+
+``` sh
+openssl req -newkey rsa:2048 -new -nodes -x509 -days 3650 -keyout key.pem -out cert.pem
```
thornjad: Is `openssl` built in to all operating systems? If not, we should mention how to install it, or provide a link, or at the very least mention that it is a requirement.

brizer:
Ok, please wait for me to modify it.

README.md
brizer: OK

README.md
brizer: Finished

README.md
brizer: Finished

README.md
brizer: Finished

README.md
brizer: Finished

README.md
thornjad: Looking back at my comment, I wasn't very clear about what I meant. I think it would be more clear to say something like:

    Then you need to run the server with `-S` for enabling SSL and `-C` for your certificate file.

    ```sh
    http-server -S -C cert.pem
    ```

    This is what should be output if successful:

    ```
    Starting up http-server, serving ./ through https
    Available on:
      https:127.0.0.1:8080
      https:192.168.1.101:8080
      https:192.168.1.104:8080
    Hit CTRL-C to stop the server
    ```

thornjad: That way, users won't need to stop to think whether they need to copy-paste the "Starting up" output

brizer: OK. It's better to understand.

PR #574, opened 2019-11-09
force to use ecstatic v3.3.2 in tests

Xmader:

force to use `ecstatic` v3.3.2 in tests

Same as https://github.com/http-party/http-server/issues/525#issuecomment-551301695, the version of  `ecstatic` used in tests is different from it when installing `http-server` globally or as a dependency. 

#525 only appears when using `ecstatic` v3.3.2 .
It is very confusing in tests.

change the version of `ecstatic` in ``package-lock.json`` to 3.3.2

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -870,9 +870,9 @@
       }
     },
     "ecstatic": {
-      "version": "3.3.0",
-      "resolved": "https://registry.npmjs.org/ecstatic/-/ecstatic-3.3.0.tgz",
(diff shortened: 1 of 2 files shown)


Xmader:
To ensure every package used in tests are the same version as installing `http-server` globally or as a dependency, just delete `package-lock.json` , or include `package-lock.json` in npm pack (https://github.com/Xmader/http-server/commit/42d7f90c60d023ea7741a35cf72b03c8791a1cb1).

thornjad:
Including the package lock in the npm package defeats the purpose of the package lock, doesn't it? In this case could we just bump the dependency in the `package.json` to `^3.3.2`?

PR #584, opened 2019-12-03
Fix basic auth options type issue

### LINKED ISSUE #583: Compliance for basic auth credentials
since `http-server` use `secure-compare` to validate basic authentication credentials and according to the predication here https://github.com/vadimdemedes/secure-compare/blob/master/index.js#L13, secure-compare will only accept string type credential arguments, this will lead to unexpected behavior with `http-server`:

if I start a server with basic authentication, let's say

```
node bin/http-server --username 1 --password 2
```

then after I visit `127.0.0.1:8080` and even I enter the correct credentials:

image

the server will respond to me with 401 since it will automatically parse `--username` and `--password` arguments to `Number` type if it can.

hence I think we need to format `options.username` and `options.password` options into `String` type before we use `secure-compare` to valid them :)

zjhmale:
Fixes #583

### DIFF
--- a/lib/http-server.js
+++ b/lib/http-server.js
@@ -108,8 +108,11 @@ function HttpServer(options) {
       // an attacker knowledge of whether the username is correct via a timing
       // attack.
       if (credentials) {
-        var usernameEqual = secureCompare(options.username, credentials.name);
-        var passwordEqual = secureCompare(options.password, credentials.pass);
+        // since the `name` and `pass` attributes of `credentials` are always string type
+        // https://github.com/DefinitelyTyped/DefinitelyTyped/blob/HEAD/types/basic-auth/index.d.ts#L15-L16
+        // so we use `.toString()` to fix https://github.com/http-party/http-server/issues/583
+        var usernameEqual = secureCompare(options.username.toString(), credentials.name);
+        var passwordEqual = secureCompare(options.password.toString(), credentials.pass);
         if (usernameEqual && passwordEqual) {
           return res.emit('next');
         }
--- a/test/http-server-test.js
+++ b/test/http-server-test.js
@@ -379,5 +379,140 @@ vows.describe('http-server').addBatch({
     teardown: function (server) {
       server.close();
     }
+  },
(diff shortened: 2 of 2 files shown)


zjhmale:
ping @thornjad

lib/http-server.js
```diff
@@ -108,8 +108,8 @@ function HttpServer(options) {
       // an attacker knowledge of whether the username is correct via a timing
       // attack.
       if (credentials) {
-        var usernameEqual = secureCompare(options.username, credentials.name);
-        var passwordEqual = secureCompare(options.password, credentials.pass);
+        var usernameEqual = secureCompare(options.username.toString(), credentials.name);
+        var passwordEqual = secureCompare(options.password.toString(), credentials.pass);
```
thornjad: Should we also convert the credentials to strings? If the server were given a numeric password, as in the tests you added, the user will likely expect numeric to work. Or are we guaranteed to receive a string as credentials?

zjhmale: from the [definition](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/HEAD/types/basic-auth/index.d.ts#L11-L17) of `basic-auth` which we use [here](https://github.com/http-party/http-server/blob/master/lib/http-server.js#L105) to parse credentials, I think the credentials are guaranteed to be string :)

thornjad: Okay, perfect, that looks like it will always be a string! Could we just add in a comment here saying the same and/or why we need to use `.toString()` here? It could be added to the comment above the `if`, or a new comment within the `if` block?

zjhmale: ping @thornjad , sorry for the late replay, the comment is added, pls review again :)

thornjad: Coule we also mention in the comment that `credentials.name` and `credentials.pass` will be strings, so we don't need to convert them to strings?

zjhmale: ping @thornjad , pls review again :)

zjhmale:
ping @thornjad

PR #591, opened 2019-12-20
Remove serverHeader

### LINKED ISSUE #237: Server exposing headers from underlying ecstatic
The 'server' header from ecstatic is exposed when running http-server which can expose potential security risks. There's currently no way to disable this from command line arguments

thornjad:
The `server` header was being served all the time by http-server to identify itself. While not usually a big deal, this is unnecessary and potentially reveals security vulnerabilities. Taking a page from engineering, I ask, why do we even need it? This PR removes the server header.

Fixes #237


thornjad:
Hey @zbynek you've been super active lately, what do you think of this PR?

zbynek (approved):
Looks good. If I understand it correctly, the API still allows setting this via `headers` option in case someone needs it.

thornjad:
Yup, so I'm not so worried about the use case that requires this header

PR #631, opened 2020-05-26
Internalize ecstatic library

### LINKED ISSUE #568: Deprecation of ecstatic
This is probably not a bug or a feature request. 
More of a _concern_. Just wanted to share with the team in case not known already.

The ecstatic package has been [deprecated](https://www.npmjs.com/package/ecstatic) and [here](https://github.com/jfhbrook/node-ecstatic/issues/259) is the feedback from the author.

This may require exploration of an alternative static file server. Any feedback is welcome, Thanks!

thornjad:
Ecstatic is [no longer maintained](), and so we cannot rely on any future updates. This PR internalizes it as the `core` module.

This PR also pulls in all of ecstatic's tests so we don't lose that coverage. Unfortunately these were written with `tap` while http-server's tests are written with `vows`. I don't have a strong preference for one over the other, but these need to be reconciled before this PR can be merged. I'm leaning toward converting to `tap` because that means rewriting only one file.

TODO:
- [x] Consolidate test files to use either `tap` or `vows`, or both.
- [ ] Ensure all tests run as expected

Fixes #568


reergymerej:
In the interim, what about a minor version bump in the dependencies to sidestep a security concern?

https://snyk.io/test/npm/ecstatic/3.3.2
> Remediation: Upgrade to ecstatic@4.1.4.

willeastcott:
The [PlayCanvas Engine repo](https://github.com/playcanvas/engine) has had a Dependabot warning about this for maybe 9 months now. I guess I could just patch the `package-lock.json`. But yeah, would be great if `http-server` bumped its `ecstatic` dependency to `4.1.3`, as @reergymerej suggests. Any objections to me submitting a PR for that?

thornjad:
Sorry I got pulled away into other life things, I need to get back to this PR. Unfortunately `ecstatic` v4 has some major incompatibilities with `http-server` (a product of the two projects being far too closely coupled for so long), so simply bumping the dependency does not work. The goal of this PR is to remove the dependency entirely, fix the vulnerability internally, and go from there. 

This PR just needs functioning tests right now if anyone has the time to take a look at them, I don't right now but will get back to it when I can.

kimegede:
Hi @thornjad, is there something you will need help with, then I will like to help out :)

zbynek:
@thornjad is this just about fixing the failing tests listed here https://travis-ci.org/github/http-party/http-server/jobs/762178038 ? If someone submitted a PR with a fix, what are the chances any maintainer would look at it?

thornjad:
Yeah tests passing is the only thing holding this up. I just haven't been able to find any time, a PR would definitely be welcomed and looked at.

thornjad:
Closing this draft in favor of #693

PR #635, opened 2020-06-10
fix: pass arguments to server.close

KoltesDigital:
- [X] The pull request is being made against the `master` branch

Rather bug fix.

As for `server.listen`, `server.close` (both for HTTP and HTTPS servers, as returned by union) takes an optional callback which is called on completion or error. This patch adds argument forwarding for `close`.

```js
	const server = createServer({
		root: path.join(__dirname, '..', 'files'),
	});

	await new Promise((resolve) => server.listen(port, resolve));
	
	...

	await new Promise((resolve) => server.close(resolve as any));
```

Well there's no tests, should I somehow write one?

As a side note, `HttpServer.prototype.close` forwards the return value of `this.server.close`, i.e. `this`. But `HttpServer.prototype.listen` does not. I think it should too.

Above code.

### DIFF
--- a/lib/http-server.js
+++ b/lib/http-server.js
@@ -206,5 +206,5 @@ HttpServer.prototype.listen = function () {
 };
 
 HttpServer.prototype.close = function () {
-  return this.server.close();
+  return this.server.close.apply(this.server, arguments);
 };


thornjad (changes requested):
I like this!

All that's left to do:

- [ ] merge `master` and let the new tests run (we've changed a lot since this PR was opened)
- [ ] add at least one test to ensure arguments are passed properly. This is less about testing your changes, more about ensuring future changes don't break this

PR #649, opened 2020-08-27
Add an -e / --errors-only flag to suppress non-error logs

Pomax:
Addresses https://github.com/http-party/http-server/issues/648

This splits the old generic `logger.info` function into separate `logger.log` and `logger.error` functions, with a new runtime flag so that folks can control whether they see "nothing", "errors only", or "all log output".

Note that the `else if(colors)` has been removed too, as the logic that makes the logging functions no-ops has been moved to being part of the declaration for logger.

Also note that the initial startup message and kill messsages will still show when errors-only is specified, as these are important runtime indicators, and their absence would make it impossible to tell whether `http-server` is running at all, something that should only be the case when the user explicit demands total log silence.

### DIFF
--- a/bin/http-server
+++ b/bin/http-server
@@ -18,15 +18,16 @@ if (argv.h || argv.help) {
     'usage: http-server [path] [options]',
     '',
     'options:',
-    '  -p --port    Port to use [8080]',
-    '  -a           Address to use [0.0.0.0]',
-    '  -d           Show directory listings [true]',
-    '  -i           Display autoIndex [true]',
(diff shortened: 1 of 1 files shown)


Pomax:
@thornjad https://github.com/http-party/http-server/graphs/contributors lists you as primary contributor, any idea who the right person is to flag for this PR? (or, if it's you: does this look good to merge?)

Pomax:
Closing this PR, I'm tired of seeing it get ignored in my open PRs list. Feel free to add this in, but I guess I'll just use my own fork from now on in the assumption this project is dead.

PR #658, opened 2020-10-12
Mention process.env.PORT in options

ts-web:
See https://github.com/http-party/http-server/issues/217

Updated the readme to document `process.env.PORT`.


thornjad (approved):
🎉

PR #682, opened 2021-04-22
Add CGI feature to enable server-side scripting

oliver-moran:

Adds a CGI feature. This enables execution of server-side scripts in a `/cgi-bin/` directory using the [Common Gateway Interface](https://en.wikipedia.org/wiki/Common_Gateway_Interface). Use case is for when small server-side scripting is needed to support static files (e.g. saving form data to a flat-file database, serving a file listing from a directory).

Added an interface to the [`cgi`](https://www.npmjs.com/package/cgi) module, with error prevention and usability assumptions (e.g. `.js` files don't need to be set to executable on Unix).

Place this file in `/public/cgi-bin/` and name it `script.js`:

``` js
console.log('Content-Type: text/plain');
console.log('');
console.log(`Hello, ${process.env.REMOTE_ADDR}!`);
```

Run `http-server` with the `--cgi` option:

``` bash
$ http-server --cgi
```

Open `http://127.0.0.1:8080/cgi-bin/script.js` and the script is executed:

``` text
Hello, 127.0.0.1!
```

Error handling, especially common mistakes that novice users of CGI would make. One known issue is in the handling of malformed headers by the `cgi` module. Error handling is not implemented in the latest released version of this module. I've added a [pull request to fix this](https://github.com/TooTallNate/node-cgi/pull/18) but the module does not appear to be maintained.

Execution of scripts on different operating systems. For JavaScript files (`.js`), there should be no difference between OSes. These don't need to be set to executable in Unix systems. On Unix, there is better choice of executing scripts in different languages (e.g. Python, bash). However, these need to be set to executable. On Windows, `.bat`, `.cmd` and `.exe` are supported in addition to JavaScript (`.js`).

It's a scripting language, so go wild!

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ This will install `http-server` globally so that it may be run from the command 
 Using `npx` you can run the script without installing it first:
 
     npx http-server [path] [options]
-     
+
 #### As a dependency in your `npm` package:
 
     npm install http-server
@@ -80,6 +80,8 @@ Using `npx` you can run the script without installing it first:
 
 `-K` or `--key` Path to ssl key file (default: `key.pem`).
 
+`--cgi` Enable CGI scripts in /cgi-bin/.
+
 `-r` or `--robots` Provide a /robots.txt (whose content defaults to `User-agent: *\nDisallow: /`)
 
 `--no-dotfiles` Do not show dotfiles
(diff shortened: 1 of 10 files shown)


thornjad:
While I think this is a really neat feature, I'm not so sure this is within the scope of http-server, which is a static content server. Would this be better suited as a companion package which depends on http-server?

thornjad:
I'm going to close this due to inactivity and possibly being out of scope for now, we can always re-open later if this is highly desired functionality

PR #459, opened 2018-08-09
add ability to pass proxyOptions 

killmenot:
http-proxy library that is used in http-server support [options](https://github.com/nodejitsu/node-http-proxy#options).

This PR adds abilities to pass proxy options via cli or programmatically (helps in cases such like https://github.com/divhide/node-grunt-http-server)

Fixes #214, fixes #246

PR #688, opened 2021-06-09
add ability to pass proxyOptions

### LINKED ISSUE #214: https certificate error when proxying to server with self-signed certificate

### LINKED ISSUE #246: Add option --auto to set proxy option autoRewrite

yannickglt:
Replaces https://github.com/http-party/http-server/pull/459

http-proxy library that is used in http-server support options.

Fixes #214, fixes #246


yannickglt:
@thornjad already made the first review of #459.

I tried to bring the changes he expected, but I didn't get that two points:
- https://github.com/http-party/http-server/pull/459#discussion_r360581849
- https://github.com/http-party/http-server/pull/459#discussion_r360583671

yannickglt:
Is it possible to be reviewed or get an advice about this please?

thornjad (changes requested):
I think this looks good! Last couple things to do:

- [ ] Add this option to `doc/http-server.1`
- [ ] Tests! They don't need to be too complex since http-proxy already has its own tests, but we should make sure the options are getting through

README.md
```diff
@@ -58,7 +58,9 @@ This will install `http-server` globally so that it may be run from the command
 |`-U` or `--utc` |Use UTC time format in log messages.| |
 |`--log-ip` |Enable logging of the client's IP address |`false` |
 |`-P` or `--proxy` |Proxies all requests which can't be resolved locally to the given url. e.g.: -P http://someurl.com | |
-|`--username` |Username for basic authentication | |
+|`--proxy-options` Pass proxy [options](https://github.com/http-party/node-http-proxy#options) using nested dotted objects. e.g.: --proxy-options.secure false
+
+`--username` |Username for basic authentication | |
```
thornjad, line 63:
```suggestion
|`--username` |Username for basic authentication | |
```

bin/http-server
```diff
@@ -62,10 +63,24 @@ var port = argv.p || argv.port || parseInt(process.env.PORT, 10),
...
 
+var proxyOptionsBooleanProps = [
+  'ws', 'xfwd', 'secure', 'toProxy', 'prependPath', 'ignorePath', 'changeOrigin',
+  'preserveHeaderKeyCase', 'followRedirects', 'selfHandleResponse'
+];
+
+if (proxyOptions) {
+  Object.keys(proxyOptions).forEach(function (key) {
+    if (proxyOptionsBooleanProps.indexOf(key) > -1) {
+      proxyOptions[key] = proxyOptions[key] === 'true';
```
thornjad, line 79: For the sake of #450, could we do 
```suggestion
      proxyOptions[key] = proxyOptions[key].toLowerCase() === 'true';
```

thornjad (changes requested):

thornjad:
Also, could you merge/rebase the current `master` in again? I don't expect conflicts, but it looks like this PR doesn't have the current tests setup, given that none are running on it

dwjohnston:
This PR solves #719

thornjad (approved):

PR #693, opened 2021-07-01
Ecstatic with tests

### LINKED ISSUE #35: Configure additional mime types

### LINKED ISSUE #276: Add --content-type command line options

### LINKED ISSUE #518: Upgrade to latest ecstatic
As of right now, ecstatic is at version 4.1.0, but we're still requiring only 3.3.0 on master. We need to make sure there are no breaking changes and upgrade, probably requiring `^4.x.x`.

This will fix #461 and #512.

### LINKED ISSUE #568: Deprecation of ecstatic

### LINKED ISSUE #689: Whether Http-Server can specify Content-Type?
**Other information (related issues, suggestions for a fix, etc):**
- Module in lib/http-server.js support config the *content-type* option, but commond in bin/http-server.js doesn't support.
- If using hs to proxy static files which ext is asp, the result will be the download. because the content-type of response is *application/octet-stream*

zbynek:

**What is the purpose of this pull request?**

Remove dependency on ecstatic -- see #631 (this is based on @thornjad 's work, only the test code was fixed here + hack from #569 replaced by a nicer fix  inside Ecstatic)

Fixes #568, fixes #518, fixes #35, fixes #276, fixes #689


zbynek:
@thornjad all green 🎉 https://travis-ci.com/github/zbynek/http-server/builds/231702777

I dropped Node 6 from the list of CI tested platforms.

thornjad:
:tada: thank you so much for the PR! I'll fit in looking over it whenever I get chances. 

For the hack from #569, do you know where the fix inside ecstatic is? I don't remember hearing anything about it, but then again, ecstatic has gone through some drama that drowned out everything else.

zbynek:
For #569 I've added the replacement into `decodePathname` directly. I *think* this way it allows you to have `\` in filenames on Linux/Mac (as long as you pass them as `%5C` in URL) and everything gets normalized to `/` on windows.

Before this change the tests were failing on Windows with infinite redirect (because the hack was only active when loading `http-server`, and the tests only load core). Now the tests are passing on Windows which suggests that the fix works, but please double check that it makes sense.

test/cli.test.js
```diff
@@ -0,0 +1,123 @@
...
+  const ecstatic = startEcstatic(options);
+
+  tearDown(ecstatic, t);
+
+  ecstatic.stdout.on('data', (msg) => {
+    checkServerIsRunning(`${defaultUrl}:${port}`, msg, t);
+  });
+});
+
+/** TODO mime types not supported in CLI
```
thornjad, line 87: I agree on commenting this out but leaving the code, but I believe the top test (using .types file) should be indirectly supported and should pass? If that's true, we can just comment out the "directly" test below.

test/core.test.js
```diff
@@ -0,0 +1,74 @@
...
+
+        if (r.type !== undefined) {
+          t.equal(
+            res.headers['content-type'].split(';')[0], r.type,
+            `content-type for \`${file}\``
+          );
+        }
+
+        if (r.body !== undefined) {
+          t.equal(eol.lf(body).trim(), r.body.trim(), `body for \`${file}\``);
```
thornjad, line 59: Why the extra trimming? Do we expect whitespace to differ?

test/express.test.js
```diff
@@ -0,0 +1,76 @@
...
+
+        if (r.type !== undefined) {
+          t.equal(
+            res.headers['content-type'].split(';')[0], r.type,
+            `content-type for \`${file}\``
+          );
+        }
+
+        if (r.body !== undefined) {
+          t.equal(eol.lf(body).trim(), r.body.trim(), `body for \`${file}\``);
```
thornjad, line 65: Same as the previous comment, why the extra trimming?

test/main.test.js
```diff
@@ -0,0 +1,286 @@
+const test = require('tap').test;
+const path = require('path');
+const fs = require('fs');
+const request = require('request');
+const httpServer = require('../lib/http-server');
+const promisify = require('util').promisify;
+
+const requestAsync = promisify(request);
+const fsReadFile = promisify(fs.readFile);
```
thornjad, line 9: What do we gain by promisify-ing rather than using the promise-based `fs` functions?

test/main.test.js
```diff
@@ -0,0 +1,286 @@
...
+          server.close();
+          resolve();
+        }
+      });
+    }),
+  ]).then(() => t.end())
+    .catch(err => {
+      t.fail(err.toString());
+      t.end();
+  });
```
thornjad, line 285: Not a problem, but why not keep the previous `.finally` approach?

test/process-env-port.test.js
```diff
@@ -0,0 +1,64 @@
...
+const request = require('request');
+const spawn = require('child_process').spawn;
+
+function getRandomInt(min, max) {
+  return Math.floor((Math.random() * ((max - min) + 1))) + min;
+}
+
+const sanePort = getRandomInt(1025, 65536);
+const floatingPointPort = 9090.86;
+const insanePorts = [-1, 65537];
```
thornjad, line 13: `-1` is a good addition, but I think we're better served by adding back `Infinity` and `'wow'` to the tests, since they capture additional edge cases

thornjad (changes requested):

test/main.test.js
zbynek: Node 8 compatibility (currently Travis is set to run Node 8,10,12,13,14)

test/main.test.js
zbynek: Node 8 compatibility again

test/process-env-port.test.js
zbynek: I wasn't able to make those work :( The ecstatic behavior under test was that invalid ports are silently ignored and default port is used.  `http-server` validates the input and dies with an exception instead. I played with some options of checking `stderr` or return code but for whatever reason those did not work for 'wow' and Infinity. I'm not sure how valuable it is to test that a process dies anyway, so I kept just the two usecases that worked.

So if we'd implement `ecstatic` behavior and just ignore invalid ports, one could easily put back the original test.

test/core.test.js
zbynek: was failing because of CRLF / LF on Windows otherwise

test/cli.test.js
zbynek: for mimetypes to work as CLI option (and hence in CLI test) some changes were needed. Note that the test would previously only pass because it did not assert anything related to the mimetype.

test/process-env-port.test.js
zbynek: @thornjad I think this is the only unresolved point now and I'm not sure if it's worth improving.

test/main.test.js
thornjad: Alright, I'm alright with that, I forgot `fs.promises` is that new. Node 8 is no longer supported but we might as well test it.

test/process-env-port.test.js
thornjad: Yeah, since http-server is meant to be used by CLI anyway (as opposed to as a library like ecstatic), I'm alright with leaving this as you've done it.

thornjad (changes requested):
I would normally complain about the `--mimetypes` addition, but I'll let that go since it's already implemented in ecstatic and we're just exposing it.

Other than that, it looks like we have a merge conflict, then I think this looks good.

thornjad:
For the conflict, I'd say just merge `master`, `rm package-lock.json` and `npm i`, that should give us a good package-lock again

thornjad:
Didn't realize I had push permissions on your fork! I just merged master in. I'll see if I can get travis hooked up again so we can see tests easily

thornjad:
Eh, looks like I don't have enough permissions to add the new Travis configuration (we were using a now-deprecated version)

thornjad:
Huh, turns out Travis is paid now. I guess I'll see what we can get from it in this PR, then move to something better before the 0.13 release.

On my latest merge, it looks like `fsevents` is doing some complaining on non-Mac hosts, do you have any insight on that? It runs fine on my local Ubuntu system, so I'm not able to repro

zbynek:
@thornjad travis is running on my branch. I changed the config to to use npm@7, it worked for all OS/Node combos except Win+Node 8: https://travis-ci.com/github/zbynek/http-server/builds/232553767

The last commit removes this combo from test matrix, I didn't manage to test it on Travis though (ran out of the free credits...), maybe you have more luck with your account.

zbynek:
@thornjad do you want to merge as is or do you want to figure out free CI first?

thornjad:
Eh, sorry yeah CI is the issue, I've run out of credits on Travis too. I'm thinking maybe yeah I'll just merge and figure out CI before releasing to npm

thornjad (approved):

PR #702, opened 2021-07-16
Replace Travis with GH Actions

thornjad:
Replaces Travis (which is now a paid service) with a GH action which tests on all the same platforms as Travis did.


thornjad:
Windows failures are an actual failing test, so I consider that a success for this PR, merging so we can see that result in master

PR #720, opened 2021-08-16
Add secure false, to allow HTTPS + proxy

dwjohnston:

Allow using the proxy feature with HTTPS

https://github.com/http-party/http-server/issues/719

Set secure flag to false for http-proxy (so it won't try verify self signed certs). 

- Question:  Is the purpose of this library for development only? If not, this probably isn't a good solution. 
- Possibly want a 'verifySslCerts' option, that you can set to false, for this use case. 

Starting a server with: 

```
http-server --proxy https://localhost:8080? ./lib -S -C localhost.crt -K localhost.key -p 8080
```

Where we are using a selfsigned certificate. 

Run the above command, check that you can access the main page at `https://localhost:8080/foobar`

I haven't written any tests yet! Working that out now.

### DIFF
--- a/lib/http-server.js
+++ b/lib/http-server.js
@@ -145,7 +145,8 @@ function HttpServer(options) {
     before.push(function (req, res) {
       proxy.web(req, res, {
         target: options.proxy,
-        changeOrigin: true
+        changeOrigin: true, 
+        secure: false, 
       }, function (err, req, res, target) {
         if (options.logFn) {
           options.logFn(req, res, {


dwjohnston:
I've tried adding tests here, but running into all sorts of issues: https://github.com/dwjohnston/http-server/tree/try-tests

thornjad:
Thank you for your PR! 

> Question: Is the purpose of this library for development only? If so, this probably isn't a good solution.

While http-server has development usage as its primary goal, we do want to be strong enough for production applications as well. So I agree, just setting it to false might be a bit heavy-handed.

It's also possible this will be covered by #688?

dwjohnston:
> It's also possible this will be covered by #688?

Definitely looks like it is covered there.

PR #728, opened 2021-09-13
open to 127.0.0.1 when host is 0.0.0.0

### LINKED ISSUE #727: Open (-o) option now defaults to http://0.0.0.0
**Steps to reproduce the issue, if applicable. Include the actual command and output and/or stack trace.**

Run the following simple command:
`http-server -o some-page.htm`

**What did you expect to happen?**
I expect http-server to open _some-page.htm_ automatically in my browser using a valid local address (http://127.0.0.1/some-page.htm)

**What actually happened?**
My browser opened to an empty page on http://0.0.0.0:8080/some-page.htm

**Tell us about your environment**
- **exact http-server version:**: 13.0.1
- **Node version:**: 14.15.2
- **Platform:**: Win 10

**Other information (related issues, suggestions for a fix, etc):**
This issue just popped up when I updated from v0.12.3 to 13.0.1 and it is caused by issue https://github.com/http-party/http-server/issues/615 and pull https://github.com/http-party/http-server/pull/642. After some research, I realized I can fix it by adding the _address to use_ `-a` to my command line arguments like so...

`http-server -o some-page.htm -a http://127.0.0.1`

But the `-o ` option previously worked without specifying the address.  Shouldn't the server still default to 127.0.1 _if_ no address is specified? As far as I can tell, all that pull request was supposed to do was to allow users to _set_ the address to 0.0.0.0, not make it as the new default. But if this was intentional, perhaps you should note in your documentation that the `-o `option should now be used with the `-a `option.

thornjad:
When using address `0.0.0.0`, open directly to `127.0.0.1`. Fixes an issue where Windows doesn't translate this address correctly.

fixes #727

### DIFF
--- a/bin/http-server
+++ b/bin/http-server
@@ -163,8 +163,7 @@ function listen(port) {
 
   var server = httpServer.createServer(options);
   server.listen(port, host, function () {
-    var canonicalHost = host,
-        protocol      = ssl ? 'https://' : 'http://';
+    var protocol = ssl ? 'https://' : 'http://';
 
     logger.info([colors.yellow('Starting up http-server, serving '),
       colors.cyan(server.root),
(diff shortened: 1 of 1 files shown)


thornjad:
Fix verified in #727

PR #733, opened 2021-10-01
make cli boolean args case insensitive

### LINKED ISSUE #450: Make Boolean command line switches case insensitive

iamsdas:

(Bug Fix)
Fixes #450 

Make the boolean args case insensitive

`http-server -d FaLse` will work properly

### DIFF
--- a/lib/http-server.js
+++ b/lib/http-server.js
@@ -54,13 +54,18 @@ function HttpServer(options) {
     options.cache === -1 ? 'no-cache, no-store, must-revalidate' :
     options.cache // in seconds.
   );
-  this.showDir = options.showDir !== 'false';
-  this.autoIndex = options.autoIndex !== 'false';
-  this.showDotfiles = options.showDotfiles;
-  this.gzip = options.gzip === true;
-  this.brotli = options.brotli === true;
+  if(options.showDir)
+    this.showDir = options.showDir.toLowerCase() !== 'false';
+  if(options.autoIndex)
+    this.autoIndex = options.autoIndex.toLowerCase() !== 'false';
+  if(options.showDotfiles)
+    this.showDotfiles = options.showDotfiles.toLowerCase() !== 'false';
+  if(options.gzip)
+    this.gzip = options.gzip.toLowerCase() !== 'false';
+  if(options.brotli)
+    this.brotli = options.brotli.toLowerCase() !== 'false';
   if (options.ext) {
-    this.ext = options.ext === true
+    this.ext = options.ext.toLowerCase() !== 'false'
       ? 'html'
       : options.ext;
   }


lib/http-server.js
```diff
@@ -54,13 +54,18 @@ function HttpServer(options) {
...
+    this.autoIndex = options.autoIndex.toLowerCase() !== 'false';
+  if(options.showDotfiles)
+    this.showDotfiles = options.showDotfiles.toLowerCase() !== 'false';
+  if(options.gzip)
+    this.gzip = options.gzip.toLowerCase() !== 'false';
+  if(options.brotli)
+    this.brotli = options.brotli.toLowerCase() !== 'false';
   if (options.ext) {
-    this.ext = options.ext === true
+    this.ext = options.ext.toLowerCase() !== 'false'
```
thornjad, line 68: Since these are actual boolean options (rather than strings), we can't `.toLowerCase()` them, and I think we can safely leave them be in this PR

lib/http-server.js
```diff
@@ -54,13 +54,18 @@ function HttpServer(options) {
...
   );
-  this.showDir = options.showDir !== 'false';
-  this.autoIndex = options.autoIndex !== 'false';
-  this.showDotfiles = options.showDotfiles;
-  this.gzip = options.gzip === true;
-  this.brotli = options.brotli === true;
+  if(options.showDir)
+    this.showDir = options.showDir.toLowerCase() !== 'false';
+  if(options.autoIndex)
+    this.autoIndex = options.autoIndex.toLowerCase() !== 'false';
```
thornjad, line 60: If we only set these values if the option is truthy, will these values be `undefined` when we use them later?

thornjad (changes requested):
In addition to these comments, please add tests!

lib/http-server.js
iamsdas: All these options are strings, so these `if conditions` will be true if the options are non empty strings. If these options are not entered then they are `undefined` by default.

lib/http-server.js
iamsdas: Actually, I checked it out (`console.log`ed the typeof options)...they are strings

lib/http-server.js
iamsdas: These were  `undefined` even before this changes (when that option is empty)

lib/http-server.js
iamsdas: for example, try `node ./bin/http-server ./public --gzip=false` in the current master branch code, this will parse gzip option as `true`

iamsdas:
So before I start with the tests, can you just confirm if the changes are ok?

lib/http-server.js
iamsdas: Now that I look at the npm test results, `Error: TypeError: options.gzip.toLowerCase is not a function` I think I need to look at it a bit more

iamsdas: Ok, found the issue:
`-g` returns a boolean
`-g false` returns a string
This is the cause of the inconsistency (adding a value after the flag)

lib/http-server.js
thornjad: If you look farther down, when we pass these options into the server, they're expected to be defined. By putting these in the `if`, they will sometimes not be defined

lib/http-server.js
thornjad: this is a feature of how the parsing library, `minimist` works, and I think we're fine leaving these as-is for now

PR #735, opened 2021-10-01
Add option for ssl ca file

### LINKED ISSUE #155: Support for CA bundles?
I have a certificate from Comodo for my domain as well as a CA bundle (several CA-related certificates). Some browsers require the CA Bundle to fully trust the website.

I'm looking for an option to pass a CA bundle to the underlying [`union` package](https://www.npmjs.com/package/union), since it [supports CA bundles](https://github.com/flatiron/union/blob/091b39a84f2f9a972ea1e15cf5922c99450c325d/README.md#createserveroptions). To make `http-server` support this, one would need to change [`bin/http-server`](https://github.com/indexzero/http-server/blob/master/bin/http-server).

mpanne:
When running the http-server with SSL enabled, there was no command line option to configure the CA file. I added this option. 

Fixes #155

### DIFF
--- a/bin/http-server
+++ b/bin/http-server
@@ -48,6 +48,7 @@ if (argv.h || argv.help) {
     '  -S --ssl     Enable https.',
     '  -C --cert    Path to ssl cert file (default: cert.pem).',
     '  -K --key     Path to ssl key file (default: key.pem).',
+    '  -A --ca      Path to ssl ca file (default: ca.pem).',
     '',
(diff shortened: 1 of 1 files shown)


thornjad (changes requested):
I love how simple this is! Besides my question below, we still need to add in:

- [ ] tests
- [ ] documentation in README.md
- [ ] documentation in doc/http-server.1

PR #736, opened 2021-10-02
Add encoding charset sniffing

### LINKED ISSUE #426: Default Encoding Charset
Thank you for inventing the server.

I am a multi-language user, Korean, and the server is responding the page with UTF-8 charset in the HEADER. However, we are still using various charset such as euc-kr.

For resolve this problem, I wrote the meta tags in the html file but it doesn't seem like working on the browser.  I think since the browser has a priority of encoding to the HEADER encoding rather than meta tags, it might necessary to set a charset in the header but I cannot find the option for that.

Would you insight me the method to add charset on the header?

Thank you in advance.

boarwell:

Resolves #426 

Added charset sniffing for text files.

To sniff charset, served files need to be read in sync (`fs.readFileSync`), rather than in stream (`fs.createReadStream`). This change would cause more memory usage and thus, if a file exceeds buffer max length (about 4GB on 64-bit architectures[^1]), Node.js will throw an error.

However, I think there is not so many cases that a plain text file is over 4GB, and this is acceptable for charset sniffing. What do you think about this?

`npm run test`

[^1]: https://nodejs.org/api/buffer.html#buffer_buffer_constants_max_length


boarwell:
@thornjad 
> What about these changes fail in Node 10? I'm not opposed to dropping support, but I don't want to without good reason

I see. That makes sense. I looked into it and found that the problem seemed to be where the buffer read into memory for sniffing was being reused to create the response stream. Commenting out the line made everything fine.

https://github.com/boarwell/http-server/blob/4392921610551087231db925d3ad65c8907e3e83/lib/core/index.js#L240

Some errors said that "Cannot set headers after they are sent to the client"[^1], so it seems there may be differences in the event loop (or something like that) between v10.x and newer versions, but I'm not sure.

So, I think we have two options:
1. (Minor Update): Stop reusing loaded buffer. It will read files two times, the first is for sniffing, the second is for creating a response.
2. (Major Update): Drop v10.x, and leave the code as-is.

I think Option 1 may be better at this time because it doesn't matter to read text files twice in most cases and it can ship this as a minor update. But also, dropping support of unmaintained versions can be a good reason for a major update. What do you think?


[^1]: https://github.com/http-party/http-server/pull/736/checks?check_run_id=3777027455

thornjad:
Node 10 dropped in master! Another merge/rebase should fix the tests!

thornjad:
Yeah those are some non-deterministic errors we're getting at random, usually re-running the tests fixes them. I've been trying to weed those out, including adding more plans. I've managed to make them better recently, but they still pop up. I reran tests for this PR and they're passing now.

But there's a conflict now

boarwell:
Fixed the conflict and passed all tests! I'm looking forward to the next release!!

PR #738, opened 2021-10-02
Added QR Code for public IP

HARDY8118:
- [ x] The pull request is being made against the `master` branch
- [ x] Tests for the changes have been added (for bug fixes / features)

Added a feature to display QR code public IP in the terminal for ease of sharing IP.

Fetched the public IP (latter one in the IP list) and used [qrcode](https://github.com/soldair/node-qrcode) package to display the qr in the terminal. 

Usability and ease of sharing IP to other devices in network.

The library provides the QR code in string format which can be used for testing/display in terminal.

### DIFF
--- a/bin/http-server
+++ b/bin/http-server
@@ -2,14 +2,16 @@
 
 'use strict';
 
-var colors     = require('colors/safe'),
-    os         = require('os'),
-    httpServer = require('../lib/http-server'),
-    portfinder = require('portfinder'),
-    opener     = require('opener'),
-    fs         = require('fs'),
-    argv       = require('minimist')(process.argv.slice(2));
+var colors = require('colors/safe'),
+  os = require('os'),
+  httpServer = require('../lib/http-server'),
+  portfinder = require('portfinder'),
+  opener = require('opener'),
+  fs = require('fs'),
+  argv = require('minimist')(process.argv.slice(2)),
+  QRCode = require('qrcode');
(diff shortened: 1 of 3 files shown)


thornjad:
This is interesting, I would never have thought of this!

Some concerns:

- I don't think this makes sense to be something that is generated by default, maybe it would be better to only generate with an option like `--qr-code`?
- I don't love that `qrcode` depends on `yargs`, which is a very heavy dependency tree. Are there any other options to make a QR code?
- Tests and documentation!
- Perhaps the biggest question of all, do we need this? It's neat, but it adds a lot of processing and code for a feature that I'm not sure will be used much? What do you think?

HARDY8118:
I really appreciate you taking time and reviewing this.
- I agree that the feature can be enabled by passing options if user wants by `--qr-code` argument or similar and I can easily make that change.
- I also understand your concern for making use of a package which brings a lot of dependencies along with it (since this package aims to provide the minamilist web server) and I think some neat and efficient way can be found for doing the same without using the `qrcode` package.
- Unfortunately I am not very sure about the testing part and how can it be done but maybe I can work on something.
- I personally use this package a lot and mostly for connecting my PC and my mobile device and it makes a lot more convenient for me by using the QR code instead of typing the whole URL. I also use my mobile device for testing and the URL generated is not static (Port number may change sometimes), I feel the need to generate the QR code everytime.

Issue #756, opened 2021-10-23
Error: Cannot set headers after they are sent to client, occurs when trying to serve index.html, only occurs in v14.0.0


abhimonk:
Getting the following error when trying to go to the webpage returned by "npx http-server" serving my index.html:

Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client

The issue does not occur if I use version 13.0.0

(See below for full stack trace)

If I repeat the reproduction steps with version 13.0.0, it works as expected.
#### Environment Versions

1. OS Type: Windows
1. Node version: 6.13.4
1. http-server version: 14.0.0

#### Steps to reproduce
1. Have just any index.html in your current folder
2. Run "npx http-server" from the current folder
3. Navigate to the URL returned (which should trigger serving your index.html)

#### Expected result
My index.html should be served and the contents of it should be displayed. This works fine in version 13.0.0.

#### Actual result

The server crashes with the following error:

[2021-10-23T17:10:29.869Z]  "GET /" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/95.0.4638.54 Safari/537.36"
(node:3956) [DEP0066] DeprecationWarning: OutgoingMessage.prototype._headers is deprecated
_http_outgoing.js:485
    throw new ERR_HTTP_HEADERS_SENT('set');

Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
    at ServerResponse.setHeader (_http_outgoing.js:485:11)
    at module.exports.ResponseStream.<computed> [as setHeader] (C:\Users\abhim\Documents\GameDev\JSDev\AttackAnimationGenerator\node_modules\union\lib\response-stream.js:100:34)
    at Object.exports.<computed> (C:\Users\abhim\Documents\GameDev\JSDev\AttackAnimationGenerator\node_modules\http-server\lib\core\status-handlers.js:57:7)   
    at Readable.<anonymous> (C:\Users\abhim\Documents\GameDev\JSDev\AttackAnimationGenerator\node_modules\http-server\lib\core\index.js:339:22)
    at Readable.emit (events.js:210:5)
    at emitErrorNT (internal/streams/destroy.js:92:8)
    at emitErrorAndCloseNT (internal/streams/destroy.js:60:3)
  code: 'ERR_HTTP_HEADERS_SENT'
}

#### Other information
[...]

Issue #634, opened 2020-06-07
Server Crashing with "Cannot set headers after they are sent to the client"


Inrixia:
**Steps to reproduce the issue, if applicable. Include the actual command and output and/or stack trace.**

Not sure sorry as there is a lot of traffic on the server, just appears to all be normal GET requests.

**What did you expect to happen?**
The server to not crash. Ideally the error would be returned as a 503.

_http_outgoing.js:485
    throw new ERR_HTTP_HEADERS_SENT('set');
    ^

Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
    at ServerResponse.setHeader (_http_outgoing.js:485:11)
    at module.exports.ResponseStream.<computed> [as setHeader] (C:\Users\Administrator\AppData\Roaming\npm\node_modules\http-server\node_modules\union\lib\response-stream.js:100:34)
    at Object.exports.<computed> (C:\Users\Administrator\AppData\Roaming\npm\node_modules\http-server\node_modules\ecstatic\lib\ecstatic\status-handlers.js:57:7)
    at ReadStream.<anonymous> (C:\Users\Administrator\AppData\Roaming\npm\node_modules\http-server\node_modules\ecstatic\lib\ecstatic.js:298:24)
    at ReadStream.emit (events.js:223:5)
    at internal/fs/streams.js:167:12
    at FSReqCallback.wrapper [as oncomplete] (fs.js:470:5) {
  code: 'ERR_HTTP_HEADERS_SENT'
}

**If the issue is a feature request, what is the motivation / use case for it?**

**Tell us about your environment**
Windows Server 2016
Node.js v12.14.1
Latest Version

**Other information (related issues, suggestions for a fix, etc):**
As far as I am aware the only thing that really needs to be done is catching the error and returning 503.

thevictor13:
This only happens with known file extensions, in my case txt, json, etc. Not sure what headers are being sent early, but without looking at the source, I'm sure it's trying to set the mime-type header, which it fails to do.

gcandrade10:
In my case it was happening for .json files, easy solution was to rollback to version 13.0.2

shrimpwagon:
Same issue. Rolling back to  13.0.2 worked

zbynek:
This is kind of a duplicate of #756 , in both cases the headers error hides some problem deeper in the code. Proposed improvement in #791. 

[...]

PR #764, opened 2021-11-14
Transfrom the buffer to string

Its-Just-Nans:
##### Relevant issues

#756 
#634

The ̀stream` is piped in `res`. Next the stream generate an error `The first argument must be one of type string or Buffer. Received type number` and launchs the `satus["500"]()` function then the `status["500"]()` function modify headers (after the piped stream), that's the crashing error appear.

To counter that, the stream musn't create error, we can :
- change `stream = Readable.from(bytes)` to `stream = Readable.from(bytes.toString())`
- remove `stream = Readable.from(bytes)` because a correct stream is created with line 334

BUT using .toString() format the buffer and a different charset can produce an error (see `toString()`in  `npm run test`).
This PR is the second (and working) option

##### Contributor checklist

- [x] Provide explanation
    - [ ] Server `--help` output
    - [ ] README.md

##### Maintainer checklist

### DIFF
--- a/lib/core/index.js
+++ b/lib/core/index.js
@@ -243,7 +243,7 @@ module.exports = function createMiddleware(_dir, _options) {
             defaultEncoding: 'UTF-8'
           });
           contentType += `; charset=${sniffedEncoding}`;
-          stream = Readable.from(bytes)
+          stream = Readable.from(bytes.toString())
         } else {
           // Assume text types are utf8
           contentType += '; charset=UTF-8';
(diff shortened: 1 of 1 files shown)


zbynek:
The idea was to reuse the stream and it works well with most supported versions of Node. Not sure how much of a performance issue it is to recreate the streamm and if it's worth fixing (#756 is not an issue with latest Node 12/14/16 and Node 10 is EOL)

PR #769, opened 2021-12-01
feat(http-server):Add a configuration file to store frequently used configuration values

wayfarer3130:
This is a simple change to be able to configure http-server via a javascript file.  This change in particular allows replacing the before methods to perform more complex configuration tasks.  It is theoretically possible to do that in nconf, but it doesn't support JavaScript out of the box, and some sort of more complex configuration is required to do things like renaming the path on the fly, or writing a non-GET handler, dynamically computing time to cache etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,8 @@ Using `npx` you can run the script without installing it first:
 
     npx http-server [path] [options]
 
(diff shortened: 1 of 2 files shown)


velzie:
There are probably security concerns to automatically executing javascript files in the same directory, especially if this is enabled by default...

PR #811, opened 2022-05-09
Fix printing interfaces on node18

alexanderankin:
Changes minor logging output issue

##### Relevant issues

Fixes #810 

##### Contributor checklist

- [x] Provide tests for the changes (unless documentation-only)
- [x] Documented any new features, CLI switches, etc. (if applicable)
    - [ ] Server `--help` output
    - [ ] README.md
    - [ ] doc/http-server.1 (use the same format as other entries)
- [x] The pull request is being made against the `master` branch

##### Maintainer checklist

- [ ] Assign a version triage tag
- [ ] Approve tests if applicable

Issue #810, opened 2022-05-09
Print interfaces in server listening message correctly on node 18


alexanderankin:
This library prints ip addresses to stdout if they are ipv4 after listening. it does this by calling [`os.networkInterfaces()`](https://nodejs.org/api/os.html#osnetworkinterfaces), which now returns `4` or `6` (unlike [before in v17](https://nodejs.org/docs/latest-v17.x/api/os.html#osnetworkinterfaces) - `family <string> Either IPv4 or IPv6`).

#### Environment Versions

1. OS Type - any
1. Node version: `$ node --version`: `v18.0.0`
1. http-server version: `$ http-server --version` present in latest master - https://github.com/http-party/http-server/blob/318c55f8c966f7a7c98c84fee3c30fce2d0bf045/bin/http-server#L229

#### Steps to reproduce

1. install node 18
2. start server

#### Expected result

server prints:

```
Starting up http-server, serving ./
Available on:
  http://127.0.0.1:8082
Hit CTRL-C to stop the server
```

#### Actual result

server prints:

```
Starting up http-server, serving ./
Available on:
Hit CTRL-C to stop the server
```

#### Other information

I think the library should take into the current value of `process.version` when printing listening interfaces. Better yet would be to not check this at all. why not show the user (properly formatted) ipv6 links? I will submit a fix that replicates the old business logic, the new business logic could be considered in a different ticket.

lygstate:
ping for this

PR #816, opened 2022-05-29
Updates details.family check for Node v18

### LINKED ISSUE #815: `Available on:` text without right hand side, where are files being served?
#### Environment Versions

1. Linux (Arch)
1. Node version: 18.2.0
1. http-server version: v14.1.0

#### Steps to reproduce

run `http-server`

#### Expected result

I expect it to tell me where the files are being served

#### Actual result

It does not share this information

```
> http-server .

Starting up http-server, serving .

http-server version: 14.1.0

http-server settings: 
CORS: disabled
Cache: 3600 seconds
Connection Timeout: 120 seconds
Directory Listings: visible
AutoIndex: visible
Serve GZIP Files: false
Serve Brotli Files: false
Default File Extension: none

Available on:
Hit CTRL-C to stop the server
```
Notice the second to last line in the above text copy-pasted from my terminal.

dmalan:
##### Relevant issues

Fixes #815.

Whereas `os.networkinterfaces` used to return a `String` value for `family` like `'IPv4'` in Node v17 and prior, per https://nodejs.org/docs/latest-v17.x/api/os.html#osnetworkinterfaces, it seems to have changed to a `Number` like `4` in Node v18, per https://nodejs.org/docs/latest-v18.x/api/os.html#osnetworkinterfaces.

##### Contributor checklist

(No existing tests to update.)

##### Maintainer checklist

### DIFF
--- a/bin/http-server
+++ b/bin/http-server
@@ -226,7 +226,7 @@ function listen(port) {
     } else {
       Object.keys(ifaces).forEach(function (dev) {
         ifaces[dev].forEach(function (details) {
-          if (details.family === 'IPv4') {
+          if (details.family === 'IPv4' || details.family === 4) {
             logger.info(('  ' + protocol + details.address + ':' + chalk.green(port.toString())));
           }
         });


zbynek:
The issue is duplicate of #810 and there is another proposed fix already in #811

thornjad:
Correct, duplicate of #811, and I _very_ marginally like #811's implementation better

PR #823, opened 2022-06-23
Close server port on exit

ngocdaothanh:
When http-server is spawned as a child process, when it exits (uppon `SIGINT`/`SIGTERM`), in certain conditions (depending on platforms and the platform conditions, it's hard to reliably reproduce), the port may still be open.

When the parent process tries to run http-server on the same port again, it will fail as the port is still open.

This PR fixes the problem by closing the port on exit.

##### Relevant issues

- https://github.com/http-party/http-server/issues/632
- https://github.com/http-party/http-server/issues/802
- https://stackoverflow.com/questions/40270182/node-child-process-doesnt-release-port-even-after-its-killed-by-parent

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/bin/http-server
+++ b/bin/http-server
@@ -242,20 +242,7 @@ function listen(port) {
       }
     }
 
-    logger.info('Hit CTRL-C to stop the server');
-    if (argv.o) {
-      const openHost = host === '0.0.0.0' ? '127.0.0.1' : host;
-      let openUrl = `${protocol}${openHost}:${port}`;
-      if (typeof argv.o === 'string') {
-        openUrl += argv.o[0] === '/' ? argv.o : '/' + argv.o;
-      }
-      logger.info('Open: ' + openUrl);
-      opener(openUrl);
-    }
-
-    // Spacing before logs
-    if (!argv.s) logger.info();
-
+    // Set up "CTRL-C" hook, before printing out "Hit CTRL-C to stop the server"
     function stopServer() {
       server.close();
       logger.info(chalk.red('http-server stopped.'));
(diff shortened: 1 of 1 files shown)


bin/http-server
```diff
@@ -255,24 +255,23 @@ function listen(port) {
...
-}
 
-if (process.platform === 'win32') {
-  require('readline').createInterface({
-    input: process.stdin,
-    output: process.stdout
-  }).on('SIGINT', function () {
-    process.emit('SIGINT');
-  });
-}
```
ngocdaothanh, line 268: Moved the block for Windows to the bottom, so that the reading order becomes natural.

bin/http-server
```diff
@@ -255,24 +255,23 @@ function listen(port) {
...
+    function stopServer() {
+      server.close();
+      logger.info(chalk.red('http-server stopped.'));
+      process.exit();
+    }
```
ngocdaothanh, line 263: Refactored from the common code below.

bin/http-server
```diff
@@ -255,24 +255,23 @@ function listen(port) {
...
-if (process.platform === 'win32') {
-  require('readline').createInterface({
-    input: process.stdin,
-    output: process.stdout
-  }).on('SIGINT', function () {
-    process.emit('SIGINT');
-  });
-}
+    function stopServer() {
+      server.close();
```
ngocdaothanh, line 260: The main purpose of this PR is to add this `server.close();` line.

ngocdaothanh:
To maintainers:
Would you please kindly release a new version, after merging this PR?

ngocdaothanh:
> When http-server is spawned as a child process, when it exits (uppon SIGINT/SIGTERM), in certain conditions (depending on platforms and the platform conditions, it's hard to reliably reproduce), the port may still be open.

I've updated the code to set up "CTRL-C" hook, before printing out "Hit CTRL-C to stop the server".

It's because the parent process relies on "Hit CTRL-C to stop the server" (or "Available on: ...") to know that http-server is ready. But at this time in the old code, the "CTRL-C" hook hasn't been set up yet. If the parent process kills http-server at this time, the port will still be open.

PR #830, opened 2022-08-26
set default listener on ipv4 and ipv6 interfaces

ssrahul96:
##### Contributor checklist

    - [ ] Server `--help` output
    - [ ] README.md

##### Maintainer checklist

Ref : https://www.grouparoo.com/blog/node-js-and-ipv6


indexzero (changes requested):
+1 this appears to be the most permissive address that supports both IPv4 & IPv6. That said, we should add `test/localhost.test.js` that validates we can connect from:

```
127.0.0.1
localhost
::1
```

KernelDeimos:
Test for this added: b7c46ac324172449770c34a07a31cb639cf4f402

merging this now

PR #836, opened 2022-10-04
Add simple implementation of serving directories as zip files

### LINKED ISSUE #798: Allow downloading directory as archive
When browsing a directory, for example one with a large number of photos or with source code for an application, I would like to download all the files in the directory. This is currently either a slow manual process of clicking each file in turn in a web browser, or requires a tool or script.

On directory listings, I would like a link to download an archive of the contents of the directory. The archive could for example be a zip file or a (plain or gzipped) tar file. The link could function similarly to the _Download ZIP_ link on GitHub repositories which builds the archive gradually and streams it to the client, so the download starts immediately.

#### Alternative Solutions
I can create an archive before using http-server and download that, but it requires me to know upfront which directories I would like to download the contents of, and for large directories, it's slow to have to package all of the contents before starting the download.

Client-side tools that can solve this is are for example `wget` with the `--spider` flag, but that's not always available, especially not on mobile devices. I have not found an easy way to do this in normal desktop and mobile browsers.

#### Additional context
I believe this can be implemented quite easily using a package like [archiver](https://www.npmjs.com/package/archiver) (streaming file creation and support for zip and tar formats) or more directly with support for just zip files using its dependency [zip-stream](https://www.npmjs.com/package/zip-stream). I'm happy to give this a go and see if I can whip up a draft pull request, assuming the extra dependency and resultant package size increase is acceptable.

I would suggest that the zip archive for a folder `/path/to/folder/` be made available at `/path/to/folder.zip`. If I file with that path already exists, the functionality could be disabled for that directory, so as not to change existing behavior. Alternatively, the archive could be available at `/path/to/folder/.zip`, which would reduce the risk of collisions.

If the functionality is implemented in such a non-disruptive way, I would also suggest that it be enabled by default when directory listings are enabled, and that it can be disabled with a command-line option. Alternatively, it could be disabled by default.

kimsey0:
This is a first draft of a new functionality that allows downloading the contents of a directory as a zip file. See the issue linked below for why I think this would be useful.

It's implemented using the [archiver](https://github.com/archiverjs/node-archiver) library, which allows streaming the compressed files directly to the response. There are currently no automated tests, but the feature can be manually tested without any additional setup by running `npm start` and visiting http://localhost:8080/img.zip.

Questions I'd like answered before I continue the work:
1. Is this something that could fit within the mission for `http-server`?
2. Are there any concerns about the number or size of the dependencies? `archiver` adds [quite a few transitive dependencies](https://www.npmjs.com/package/archiver?activeTab=dependencies). The number could probably be reduced by using a lower-level library like [zip-stream](https://www.npmjs.com/package/zip-stream) at the cost of more complex wrapping code.
3. How should this be configured? I've currently added a `--serve-dir-zip` command line argument that controls it. Should it be enabled by default?
4. What happens if there's a zip file with the same name as a directory? Right now, it's served instead of zipping the directory. The risk of collisions could potentially be reduced by serving directory zips at `/directory/.zip` instead of `/directory.zip`.
5. Should other archive formats be supported? Support for tar files, optionally gzipped, is included `archiver`, so it's trivial to enable.
6. And as a last minor question, should a link to the zip file be added to directory listings, and in that case where?

##### Relevant issues
Fixes #798.

##### Contributor checklist

    - [x] Server `--help` output
    - [ ] README.md

##### Maintainer checklist

### DIFF
--- a/lib/core/index.js
+++ b/lib/core/index.js
@@ -15,6 +15,7 @@ const status = require('./status-handlers');
 const generateEtag = require('./etag');
 const optsParser = require('./opts');
 const htmlEncodingSniffer = require('html-encoding-sniffer');
+const serveDirZip = require('./serve-dir-zip');
 
 let httpServerCore = null;
 
@@ -352,6 +353,9 @@ module.exports = function createMiddleware(_dir, _options) {
               // This means we're already trying ./404.html and can not find it.
               // So send plain text response with 404 status code
               status[404](res, next);
+            } else if (path.extname(parsed.pathname) === '.zip' && opts.serveDirZip) {
+              // This is a request for a zip file that doesn't exist. Try to serve the directory as a zip file instead.
+              serveDirZip(opts, stat)(req, res);
             } else if (!path.extname(parsed.pathname).length && defaultExt) {
               // If there is no file extension in the path and we have a default
               // extension try filename and default extension combination before rendering 404.html.
--- a/bin/http-server
+++ b/bin/http-server
@@ -60,7 +60,8 @@ if (argv.h || argv.help) {
     '  --no-dotfiles      Do not show dotfiles',
     '  --mimetypes        Path to a .types file for custom mimetype definition',
     '  -h --help          Print this list and exit.',
-    '  -v --version       Print the version and exit.'
+    '  -v --version       Print the version and exit.',
+    '  --serve-dir-zip    Serve zip file of directory contents [true]'
   ].join('\n'));
   process.exit();
 }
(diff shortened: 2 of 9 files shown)


kimsey0:
@thornjad, perhaps you can help clear some of these up when you have time and energy? (In any case, thanks for this great utility. 😄)

lib/core/index.js
```diff
@@ -352,6 +353,9 @@ module.exports = function createMiddleware(_dir, _options) {
               // This means we're already trying ./404.html and can not find it.
               // So send plain text response with 404 status code
               status[404](res, next);
+            } else if (path.extname(parsed.pathname) === '.zip' && opts.serveDirZip) {
```
kimsey0, line 356: This should probably be case insensitive.

kimsey0:
@thornjad, any chance you could answer a few of the questions above? :smile:

kimsey0:
@thornjad, any input on whether this could conceivably merge in the right state?

kimsey0:
@thornjad, I still think this would be useful. Any chance you think so too?

kimsey0:
Still relevant.

PR #838, opened 2022-10-27
added code for creating robots.txt

### LINKED ISSUE #825: robots.txt flag is not working
passing "-r true" or "--robots true" on command line.
possible reasons:
- http-server.js, near line 113, there is "options.robots === true", but in fact the value is a string, so === is not right.
- if true, looks like returns "user-agent:" string as body, but shouldn't it call reset.setHeader or something?

rskbansal:
### Description 
Previously, the code was written so as to return text data when [localhost/robots.txt](http://localhost/robots.txt) is visited by providing the `-r` or the `--robots` argument. Now, if the above mentioned arguments are passed, it creates a robots.txt file in the root directory of the server (either `./` or `./public`).
### Relevant issues

Fixes #825 

### Contributor checklist

- [x] Provide tests for the changes (unless documentation-only)
- [x] Documented any new features, CLI switches, etc. (if applicable)
    - [x] Server `--help` output
    - [x] README.md
    - [x] doc/http-server.1 (use the same format as other entries)

### Maintainer checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ This will install `http-server` globally so that it may be run from the command 
 |`-S`, `--tls` or `--ssl` |Enable secure request serving with TLS/SSL (HTTPS)|`false`|
 |`-C` or `--cert` |Path to ssl cert file |`cert.pem` | 
 |`-K` or `--key` |Path to ssl key file |`key.pem` |
-|`-r` or `--robots` | Automatically provide a /robots.txt (The content of which defaults to `User-agent: *\nDisallow: /`)  | `false` |
+|`-r` or `--robots` | Automatically creates a /robots.txt file (The content of which defaults to `User-agent: *\nDisallow: /`)  | `false` |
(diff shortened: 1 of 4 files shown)


rskbansal:
@thornjad

BigBlueHat:
@rskbansal not sure we want an http-server writing files. What's the motivation for the change?

rskbansal:
A `robots.txt` file shall be physically present on the server, that's what it is meant to be

zbynek:
> file shall be physically present 

but why? For the user it's only important that the response for [localhost/robots.txt](http://localhost/robots.txt) contains the data, not that the file exists. The current implementation looks fine (except for the `=== true` check mentioned in #825)

BigBlueHat:
> A `robots.txt` file shall be physically present on the server, that's what it is meant to be

If the file needs to be physically present for your use case, I'd recommend generating it prior to hosting and just not using the `--robots` option.

@zbynek thinks for connecting to #825. That does need a fix.

PR #852, opened 2023-02-06
Serve compressed files if already exist

### LINKED ISSUE #780: Missing HTTP Response Header "Content-Encoding: gzip"
-g option will enable content-type: application/gzip; charset=utf-8
but this is not enough, it has to have "Content-Encoding: gzip" also in response header
here is an example of error message when attempting to serve gzip content:
image

### LINKED ISSUE #844: Fails to serve Unity/Brotli
Exporting a WebGL build from unity and then running `http-server -o --cors --brotli`

I see `Serve Brotli Files` is true, but I get an error where files like `Build/world.framework.js.br` are not being served with `Content-Encoding: br` in the response header.

Ultimately this means Unity WebGL builds are not viewable with `http-server`

#### Environment Versions

1. OS Type: OSX
1. Node version: `v16.13.1`
1. http-server version: `v14.1.1`

De-Panther:
If file is already compressed (ends with `.gz` or `.br`), no need to add those file extensions.
Sometime files are already compressed and needs to serve with the correct header.
In some cases engines like Unity output some of the files as compressed and others not.

This change allows to serve already compressed files even if `Serve GZIP Files` or `Serve Brotli Files` are false.

I'm not sure how to write a test, or if this change should be documented.
It seems to me like an obvious outcome when serving a `.gz` or `.br` file. So I think it's a bug fix rather than a feature or a change in a feature.

##### Relevant issues

Resolves #844
Resolves #780

##### Contributor checklist

    - [ ] Server `--help` output
    - [ ] README.md

##### Maintainer checklist

### DIFF
--- a/lib/core/index.js
+++ b/lib/core/index.js
@@ -201,8 +201,8 @@ module.exports = function createMiddleware(_dir, _options) {
       )
     );
     // determine compressed forms if they were to exist
-    gzippedFile = `${file}.gz`;
-    brotliFile = `${file}.br`;
+    gzippedFile = file.endsWith('.gz') ? file : `${file}.gz`;
+    brotliFile = file.endsWith('.br') ? file : `${file}.br`;
 
     Object.keys(headers).forEach((key) => {
       res.setHeader(key, headers[key]);


grybykm (approved):

De-Panther:
Apparently there were already tests for `gzip` and `brotli`, and my first commit broke the `Content-Type` tests.
I moved the `defaultType` to the start of the process right after the options are parsed.
Also moved the MIME lookup to right after we get the file, which made redundant the special lookup for compressed files.

Some unrelated CLI tests don't work for me both on master and on this PR.

@thornjad is there something else I should do for this PR to be approved?
Thanks

PR #885, opened 2024-08-10
Added option for custom headers

### LINKED ISSUE #884: User-provided custom headers

In a project I was working on, I wished to use CORP headers in order to use SharedArrayBuffer (which is only possible if we don't allow cross origin requests through the CORP headers, which are different from the CORS headers already available in the http-server script).

The solution I propose would be to add a CLI argument to provide a list of headers, potentially a string containing a semi-colon separated list of "header:value" entries.

#### Alternative Solutions

I used to use a python library to run an HTTPS server to test my application, but the library doesn't seem to be available on my distribution anymore. And this application seems to be more user-friendly anyway, so it would be best if we upgraded this module with the option to add custom headers.

#### Additional context

N/A

kipcode66:
This PR adds an option to provide custom headers.

##### Relevant issues

Resolves #884

##### Contributor checklist

- [X] Documented any new features, CLI switches, etc. (if applicable)
    - [X] Server `--help` output
    - [X] README.md
    - [X] doc/http-server.1 (use the same format as other entries)
- [X] The pull request is being made against the `master` branch

##### Maintainer checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,7 @@ This will install `http-server` globally so that it may be run from the command 
 |`-r` or `--robots` | Automatically provide a /robots.txt (The content of which defaults to `User-agent: *\nDisallow: /`)  | `false` |
 |`--no-dotfiles` |Do not show dotfiles| |
 |`--mimetypes` |Path to a .types file for custom mimetype definition| |
+|`-H` or `--headers` |Semi-colon separated list of headers to use (Ex.: `Cross-Origin-Resource-Policy:same-site;Cross-Origin-Opener-Policy:same-origin`)| |
 |`-h` or `--help` |Print this list and exit. |   |
 |`-v` or `--version`|Print the version and exit. | |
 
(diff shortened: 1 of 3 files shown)


kipcode66:
I was coming back to make tests for this PR and fix the fact it doesn't handle colons in the header values, but I saw there was an other PR https://github.com/http-party/http-server/pull/887 which uses the curl approach to adding headers. Which approach would be prefered? should I just close this one if we'd rather use the curl-based one?

KernelDeimos:
> I was coming back to make tests for this PR and fix the fact it doesn't handle colons in the header values, but I saw there was an other PR #887 which uses the curl approach to adding headers. Which approach would be prefered? should I just close this one if we'd rather use the curl-based one?

Thanks for bringing the dupe to my attention. I like the idea of using known conventions where possible. Let's move focus to #887, and I've also opened discussion #899 so we can talk about superset syntax (i.e. maybe we allow semicolon-separated values as well if it doesn't conflict with curl's convention, because that seems more convenient to me).

PR #901, opened 2024-10-15
Proxy websockets

pft:
Proxy websockets when --websocket is given.

##### Relevant issues

Fixes issue https://github.com/http-party/http-server/issues/142

I took the PR from @fabioimpe at https://github.com/http-party/http-server/pull/528, removed style changes (and made sure not to add any of my own), and removed the changes to the changeOrigin parameter.

And of course, I tested whether it worked, which it does.


KernelDeimos:
Is there any way to test this? This should have tests before I merge it, if a unit test for this is complex I'll do a manual test later.

pft:
Well, it relies on another server working with websockets. A unit test could be done, but would require quite a bit of work.

pft:
Thank you!

PR #905, opened 2024-11-04
Fix Node.js CI workflow

### LINKED ISSUE #904: Node.js CI workflow is broken
It seems like bumping `tap` to `tap@^21.0.1` has broken CI workflow:
- it requires Node >=20
- it no longer support `spec` reporter
- it returns exit code 1 if coverage is not 100%

#### Environment Versions

1. Any OS
1. All available Node versions
1. http-server version: latest

#### Steps to reproduce

1. Make a PR and see Node.js CI workflow in `Actions` tab

#### Expected result

1. Green jobs
2. Green build badge

#### Actual result

Here's workflow ran against current master branch:
workflow ran against current master branch

ixth:
It seems like upgrading `tap` to `21.x.x` has caused tests to fail on most Node versions in the CI. `tap 21` only supports `node>=20`, so tests fail when run on older versions, also test coverage and reporter settings hanged, so `tap` always returns exit code `1`.

Sorting out all this, I also ran into several other issues:

- Node versions 12 and 14, which are used in tests, are significantly outdated, causing `setup-node` to struggle with locating compatible images. At least, now it shows following error: `Unable to find Node version '12.x' for platform darwin and architecture arm64.`
- `checkout@v2` and `setup-node@v2` are outdated and still rely on Node 12.

To resolve all this I did following
- Downgraded to `tap@20`, which still supports Node 18
- It seems like `spec` reporter is not available anymore, so I changed it for `terse`
- Added `--allow-incomplete-coverage` for `tap` not to return exit code 1
- Dropped `12.x` and `14.x` node versions from  tests job matrix and added `20.x`, `21.x`, and `22.x` to it
- Bumped `engines.node` in package.json to the version CI tests agains in `16.x` scenario

##### Relevant issues

Fixes #904 

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -21,9 +21,9 @@ jobs:
         os: [ubuntu-latest, macOS-latest, windows-latest]
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4.2.2
     - name: Use Node.js ${{ matrix.node-version }} on ${{ matrix.os }}
-      uses: actions/setup-node@v2
+      uses: actions/setup-node@v4.1.0
       with:
         node-version: ${{ matrix.node-version }}
         cache: 'npm'
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


KernelDeimos:
Fantastic, thanks! You got all the nuances to fix this perfectly.

Issue #906, opened 2024-11-08
http-server can be configured with more flexible settings, especially regarding proxy functionality


JiayangShen:
#### What's the problem this feature will solve?
http-server now has the capability to configure proxy functions; however, this feature is quite basic and can only forward all unhandled HTTP requests to a single address. Often, we want to be able to configure multiple proxies or have more flexible and complex proxy settings. For example, requests to /api/xxx should be proxied to http://192.168.6.8:8080, and requests to /api2/xxx should be proxied to http://192.168.6.8:8082.
#### Describe the solution you'd like
Configuring multiple and complex proxies with http-server using command-line options can be challenging. We can conveniently achieve this functionality by using a configuration file (such as http-server.json or http-server.config.js).

For example, an http-server.json configuration file might look like this:
```json
{
    "port": 8092,
    "proxy": {
        "/api/": {
            "target": "http://192.168.6.8:8080",
            "changeOrigin": true
        },
        "/api2/": {
            "target": "http://192.168.6.8:8082",
            "changeOrigin": false
        }
    }
}
```
When http-server runs, it checks for the presence of this default-named configuration file in the current directory. If it exists, http-server reads the configuration and performs the corresponding functions based on the settings.

i3ene:
It would be nice if the configuration could be specified with `--proxy-config <path/to/my-http-config>.json`.

PR #907, opened 2024-11-14
added proxy config for complex proxy rules

i3ene:
### Description
Added an additional `--proxy-config` parameter that takes in a `.json` file (with path).
The use of it overrides any other proxy configuration! (Well, it actually disables the other options, not overrides it, but this could also be implemented.)
This allows to write individual rules like:
```json
{
  "/api/*": {
    "target": "localhost:8090",
    "changeOrigin": true,
    "secure": false,
    "ws": false,
    "pathRewrite": {
      "^/api": ""
    }
  },
  "*": {
    "target": "localhost:8070",
    "changeOrigin": false
  }
}
```

### Relevant Issues
This addresses #439, #901 and #906.
I wanted to make a first approach with this to maybe incorporate more functionality different people need. It does rely on a file and therefor isn't CLI only configurable, but with the complexity some rules can have, I think it is not feasible.

### Tasks
There are still a few things to do:
 - [ ] Write tests
 - [ ] Create documentation
 - [ ] Default proxy (use default target and port) behavior if not rule matches

### DIFF
--- a/bin/http-server
+++ b/bin/http-server
@@ -51,6 +51,7 @@ if (argv.h || argv.help) {
     '',
     '  -P --proxy       Fallback proxy if the request cannot be resolved. e.g.: http://someurl.com',
     '  --proxy-options  Pass options to proxy using nested dotted objects. e.g.: --proxy-options.secure false',
+    '  --proxy-config   Pass in .json configuration file. e.g.: ./path/to/config.json',
     '',
     '  --username   Username for basic authentication [none]',
     '               Can also be specified with the env variable NODE_HTTP_SERVER_USERNAME',
@@ -76,6 +77,7 @@ var port = argv.p || argv.port || parseInt(process.env.PORT, 10),
     sslPassphrase = process.env.NODE_HTTP_SERVER_SSL_PASSPHRASE,
     proxy = argv.P || argv.proxy,
     proxyOptions = argv['proxy-options'],
+    proxyConfig = argv['proxy-config'],
     utc = argv.U || argv.utc,
     version = argv.v || argv.version,
     baseDir = argv['base-dir'],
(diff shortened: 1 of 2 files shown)


KernelDeimos:
Hi, this has been active for a bit but it needs tests. If you're still interested in having this merged that would be a great help. In either case I'll try to come back to this in a few days and give it a proper review, maybe I'll add the test myself.

i3ene:
Sorry for not responding so long, I'll try to incorporate some tests in the next few days! I'm still behind this functionality and think it would help a lot of others.

i3ene:
__Some discussion:__
Should the rules be written as real Regex, or should we use simplified patterns?
At the moment I try to simplify it as to make it more usable. For example:
| Simplified | Actual Regex | |
| --- | --- | --- |
| `/some/path/*` | `/some/path/.*` | |
| `/some/$/path` | `/some/[^/]*/path` | (not implemented yet) |

KernelDeimos:
I see the problem you're solving there. I think it should be either just regex or a convention that's common for paths in a well-known platform or framework. Maybe glob syntax?

KernelDeimos:
This will probably be merged after #933

KernelDeimos:
Conflicts need to be fixed. Also it looks like some thought was given [in another thread](https://github.com/http-party/http-server/pull/365#issuecomment-307281707) to the syntax for rules like the ones you brought up, just for URL rewrites instead of proxying. Firebase config was mentioned, and it looks like it uses the simplified/glob syntax.

bin/http-server
```diff
@@ -254,6 +258,12 @@ function listen(port) {
     proxyOptions = undefined;
   }
 
+  // TODO: handle this in respect of proxyConfig
```
i3ene: I am not sure how to handle this. Any idea? I think it would not be possible to have `proxyAll` and `proxyConfig` work at the same time as they solve different problems.

KernelDeimos: I agree, I think the best solution is have `--proxy-all` throw an error when using a config, since the config is already route-based.

Alternatively, a useful behavior for this could be as a check that there's a route for `"*"` and an error if there isn't.

KernelDeimos:
Nice, `minimatch` looks like a good choice! Merging now (actually waiting for CI tests, for some reason they didn't run yet)

i3ene:
I didn't have much time recently, but will look into this tomorrow.

PR #931, opened 2025-10-10
doc(upgrade) ask our intern claude how they might write an upgrade guide

indexzero:
@jelveh follow-up from our email chain with a decent (but very wordy and very `claude` sounding) Upgrade Guide for `15.0.0`

Excited to see this release go live 🐢 🚀 ✨

### DIFF
--- a/doc/UPGRADE.md
+++ b/doc/UPGRADE.md
@@ -1,0 +1,428 @@
+# Upgrade Guide: http-server v14.1.1 → v15.0.0
+
+> **WARNING: Major release with breaking changes after 3 years**
+>
+> **Test thoroughly before production deployment**
+
+## Critical Breaking Changes
+
+### 1. Node.js Version Requirement
+- **Minimum**: Node.js 16.20.2
+- **Impact**: Server will not start on older versions
+- **Action**: Upgrade Node.js before upgrading http-server
+
+The specific patch version (16.20.2) is unusual. Verify this requirement hasn't changed.
+
+### 2. IPv6 Default Binding - SECURITY CRITICAL
+
+**This is the most dangerous change in this release.**
+
+- **Old**: Default host `0.0.0.0` (IPv4 only)
+- **New**: Default host `::` (dual-stack IPv4+IPv6)
+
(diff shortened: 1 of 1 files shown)


doc/UPGRADE.md
```diff
@@ -0,0 +1,428 @@
...
+- **New**: TAP v21
+- **Impact**: Test framework completely changed
+
+## Security Warnings
+
+### Header Injection Vulnerability (New -H Flag)
+
+**SEVERITY: CRITICAL**
+
+The new `-H/--header` option has NO CRLF validation. This allows HTTP response splitting attacks if you pass untrusted input.
```
KernelDeimos: The first statement, that that `-H/--header` option has no CRLF validation, is true. The second statement that this allows injecting CRLF characters into headers, is **not** true. `union` is already protecting against this:

```
TypeError [ERR_INVALID_HTTP_TOKEN]: Header name must be a valid HTTP token ["X-Custom: MyValue
X-Antoher: OtherValue"]
    at ServerResponse.setHeader (node:_http_outgoing:701:3)
    at new module.exports (/home/edube/repos/http-server/node_modules/union/lib/response-stream.js:31:21)
    at new module.exports (/home/edube/repos/http-server/node_modules/union/lib/routing-stream.js:30:17)
    at Server.requestHandler (/home/edube/repos/http-server/node_modules/union/lib/core.js:27:25)
    at Server.emit (node:events:507:28)
    at parserOnIncoming (node:_http_server:1153:12)
    at HTTPParser.parserOnHeadersComplete (node:_http_common:117:17) {
  code: 'ERR_INVALID_HTTP_TOKEN'
}
```

When #932 is merged `http-server` will throw an error immediately when the configuration is invalid instead of waiting until attempting to send the first response.

doc/UPGRADE.md
```diff
@@ -0,0 +1,428 @@
...
+
+**Only use --cors for**:
+- Public CDN resources (fonts, images, libraries)
+- Open APIs with no authentication
+- Development/testing (never production)
+
+**Never use --cors with**:
+- APIs with authentication
+- Any endpoint serving user-specific data
+- Internal resources
```
KernelDeimos: As an upgrade guide, should this be here? This won't cause a regression for anyone who isn't already using the flag. Maybe this belongs in the documentation for this flag.

PR #953, opened 2026-02-01
implement --allowed-hosts flag

### LINKED ISSUE #758: Host header check

Defend against DNS rebinding attacks (https://en.wikipedia.org/wiki/DNS_rebinding). 

Add a new option for specifying the permitted value of the Host HTTP header (i.e. the hostname of the server to which the
request is sent).

A key part of the DNS rebinding attack is that the user's browser will be sending a request to an internal server with an invalid Host header for that server, i.e., the Host header's value is not a real DNS name for the IP address on which the server is running.  If the server checks the Host header against the value provided in the new option, the attack fails.  For an example of a server that does this, see https://angular.io/cli/serve and its "--host" option.

#### Alternative Solutions

I don't see any other solution other than to use a different package.

#### Additional context

Note that DNS rebinding is an issue even when running the server on a development machine behind a firewall.  Any one who has access to the develop machine server who also browses to a malicious website can open up access to that development server.

velzie:
The --allowed-hosts flag allows users to block requests with `Host` headers that aren't whitelisted, preventing against [DNS rebinding](https://en.wikipedia.org/wiki/DNS_rebinding) attacks

##### Relevant issues
Fixes #758 

##### Contributor checklist

- [x] Provide tests for the changes (unless documentation-only)
- [x] Documented any new features, CLI switches, etc. (if applicable)
    - [x] Server `--help` output
    - [x] README.md
    - [x] doc/http-server.1 (use the same format as other entries)

##### Maintainer checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,8 +93,9 @@ with the provided Dockerfile.
 |`-K` or `--key` |Path to ssl key file |`key.pem` |
 |`-r` or `--robots` | Automatically provide a /robots.txt (The content of which defaults to `User-agent: *\nDisallow: /`)  | `false` |
 |`--no-dotfiles` |Do not show dotfiles| |
-|`--hide-permissions` |Do not show file permissions| |
 |`--mimetypes` |Path to a .types file for custom mimetype definition| |
+|`--hide-permissions` |Do not show file permissions| |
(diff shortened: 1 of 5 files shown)


velzie:
We might want to add more validation to the argument - I can see users passing in something like `--allowed-hosts=localhost:8080` which would break, but maybe we should leave that until the options refactor

PR #954, opened 2026-02-09
Allow setting changeOrigin to false

### LINKED ISSUE #831: Cannot override changeOrigin proxy option
I am attempting to authenticate to a web application using SSO. The apache web server runs on localhost:8080, while the http-server runs on localhost:8443. The problem I'm experiencing is that http-server will rewrite the `host` header as localhost:8080. As a result, I can authenticate with the SSO server successfully, but I am redirected back to my web server at localhost:8080.

I expected to solve this problem by setting the changeOrigin option to false, but it isn't working as I'd expect it to work.

#### Environment Versions

1. OS: Ubuntu 18
1. Node version: 14.17.2
1. http-server version: 14.1.1

#### Steps to reproduce
Run http-server with `--proxy-options.changeOrigin false`.

#### Expected result
The `changeOrigin` flag is true by default, but can be overridden with the above option.

#### Actual result
The `changeOrigin` flag is always set to true and the option is ignored.
See lib/http-server.js:
```js
proxy.web(req, res, {
        target: options.proxy,
        changeOrigin: true
}
```

#### Other information

I will open a PR if this behaviour is deemed to be a bug.

JensOlesen47:
If changeOrigin is set to false in the proxyOptions, it will now be honoured.

##### Relevant issues

Fixes #831.

##### Contributor checklist

    - [ ] Server `--help` output
    - [ ] README.md

##### Maintainer checklist

### DIFF
--- a/lib/http-server.js
+++ b/lib/http-server.js
@@ -167,13 +167,13 @@ function HttpServer(options) {
         if (!minimatch(req.url, key)) continue;
         req.proxy ??= {};
         var matchConfig = options.proxyConfig[key];
-        
+
         if (matchConfig.pathRewrite) {
           Object.entries(matchConfig.pathRewrite).forEach(rewrite => {
             req.url = req.url.replace(new RegExp(rewrite[0]), rewrite[1]);
           });
         }
-        
+
         var configEntries = Object.entries(matchConfig).filter(entry => entry[0] !== "pathRewrite");
         configEntries.forEach(entry => req.proxy[entry[0]] = entry[1]);
         break;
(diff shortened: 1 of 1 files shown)


KernelDeimos:
Thanks, merging this as-is for the moment, but also I'd like to add a note here: do we need to also change this line? (this affects websockets)
https://github.com/http-party/http-server/blob/f80aca09e9c9f265bc65c37f8d681ff585ec206d/lib/http-server.js#L271

JensOlesen47:
Thanks Eric,

> Thanks, merging this as-is for the moment, but also I'd like to add a note here: do we need to also change this line? (this affects websockets)
> 
> https://github.com/http-party/http-server/blob/f80aca09e9c9f265bc65c37f8d681ff585ec206d/lib/http-server.js#L271

Good catch! I would assume that needs to be changed as well. I could open another PR for it, but I don't have much / any experience with websockets so I'm not sure how I'd test it.

KernelDeimos:
> Thanks Eric,
> 
> > Thanks, merging this as-is for the moment, but also I'd like to add a note here: do we need to also change this line? (this affects websockets)
> > https://github.com/http-party/http-server/blob/f80aca09e9c9f265bc65c37f8d681ff585ec206d/lib/http-server.js#L271
> 
> Good catch! I would assume that needs to be changed as well. I could open another PR for it, but I don't have much / any experience with websockets so I'm not sure how I'd test it.

No problem, I'll open a separate issue for this one.

Issue #949, opened 2026-01-29
ReferenceError: colors and logger warning are not defined when using --websocket flag


HarshVz:
## Bug: `colors is not defined` when using `--websocket` flag

When attempting to use the `--websocket` flag without the `--proxy` flag, the server crashes with a `ReferenceError: colors is not defined` error.

#### Environment Versions

1. **OS Type**: Windows 11
2. **Node version**: `v22.17.0` (or run `node --version`)
3. **http-server version**: `v14.1.2` (master branch, unreleased)

#### Steps to reproduce

1. Clone the repository and install dependencies:

   ```bash
   git clone https://github.com/http-party/http-server.git
   cd http-server
   npm install
   ```

2. Run http-server with the `--websocket` flag but without `--proxy`:

   ```bash
   node bin/http-server --websocket
   ```

3. Observe the crash

#### Expected result

The server should start and display a warning message:

```
WebSocket proxy will not be enabled because proxy is not enabled
```

Then continue running normally.

#### Actual result

The server crashes immediately with:

```
C:\Users\Harsh\V0 Builds\express\http-server\bin\http-server:233
      logger.warning(colors.yellow('WebSocket proxy will not be enabled because proxy is not enabled'));
                     ^

ReferenceError: colors is not defined
    at listen (C:\Users\Harsh\V0 Builds\express\http-server\bin\http-server:233:22)
    at C:\Users\Harsh\V0 Builds\express\http-server\bin\http-server:185:5
    ...
```

also, warning is not defined as well for logger:

<img width="341" height="30" alt="Image" src="https://github.com/user-attachments/assets/dc12cc03-1a61-4ccf-b6e8-f0ddd0244e29" />

#### Root Cause

There are **two bugs** on line 233 in `bin/http-server`:

1. **Undefined variable**: The code references `colors` which is never imported. The file imports `chalk` instead (line 5).
2. **Undefined method**: The code calls `logger.warning()` which doesn't exist. The logger object only has `info` and `request` methods (defined at lines 138-171).

**Current code (line 233):**

```javascript
logger.warning(
  colors.yellow(
    "WebSocket proxy will not be enabled because proxy is not enabled",
  ),
);
```

**Should be:**

```javascript
logger.info(
  chalk.yellow(
    "WebSocket proxy will not be enabled because proxy is not enabled",
  ),
);
```

#### Additional Bug: Same issue in error handler

[...]

PR #958, opened 2026-02-27
Replace chalk with picocolors

roli-lpci:
## Summary

- Replace `chalk` (v4, 5 transitive deps) with `picocolors` (v1, zero deps)
- Fix `ReferenceError: colors is not defined` in the `--no-panic` error handler
- Fix `TypeError: logger.warning is not a function` in the WebSocket proxy check

## Changes

**bin/http-server**
- `require('chalk')` → `require('picocolors')`
- Fix undefined `colors` variable references (lines 115-118) that caused crashes
  in the `--no-panic` error handler — `colors` was never imported, should have
  been `chalk` (now `pc`)
- Fix `logger.warning()` call — the logger object has no `.warning()` method;
  changed to `logger.info()` which is the available output method

**package.json**
- Remove `chalk` ^4.1.2
- Add `picocolors` ^1.1.1

All color methods used (`red`, `cyan`, `yellow`, `green`, `magenta`, `bold`)
have exact equivalents in picocolors. Both libraries respect `NO_COLOR` and
`FORCE_COLOR` environment variables.

## Testing

```
Asserts:  528 pass  0 fail  528 of 528 complete
Suites:    41 pass  0 fail    41 of 41 complete
```

Addresses #949


KernelDeimos:
picocolors seems cool - 2x faster than chalk is great, 14x smaller is even better - but what issues are we going to introduce with the smaller alternative? Is chalk 14x the size for a reason? I noticed there's an [included library that reports color support for known terminals](https://github.com/chalk/chalk/blob/main/source/vendor/supports-color/index.js) - I reckon we might need that.

Btw we might [repeat the same mistake that caused #949](https://github.com/http-party/http-server/issues/949#issuecomment-4049141311); please use the name `chalk` instead of `pc` even though the `picocolors` package is being used.

roli-lpci:
Good question. picocolors does handle color support detection -- it checks NO_COLOR, FORCE_COLOR, and whether stdout is a TTY. The difference is chalk bundles a larger lookup table that maps specific terminal app names (iTerm, ConEmu, etc.) to 256-color and truecolor support levels.

But http-server only uses basic colors -- red, green, cyan, bold -- never 256-color or truecolor. So the simpler detection in picocolors covers everything we actually need here. No functionality lost for this use case.

